PR #11, opened 2022-07-08
INTYGFV-14568: Added sms and email notification functionality

mikeyOne:


customer-terminate-service/app/src/main/java/se/inera/intyg/cts/infrastructure/service/SendPackageNotification.java
```diff
@@ -0,0 +1,158 @@
...
+import org.springframework.transaction.annotation.Transactional;
+import se.inera.intyg.cts.domain.model.Termination;
+import se.inera.intyg.cts.domain.model.TerminationId;
+import se.inera.intyg.cts.domain.repository.TerminationRepository;
+import se.inera.intyg.cts.domain.service.SendNotification;
+import se.inera.intyg.cts.infrastructure.integration.SendEmail;
+import se.inera.intyg.cts.infrastructure.integration.SendSMS;
+
+@Service
+public class SendPackageNotification implements SendNotification {
```
mww-sopra-steria, line 16: I think you should move any changes to the domain object out of this service. This service purpose is to send notifications. It should leave it to the domain to make necessary updates to the domain objects itself.

customer-terminate-service/app/src/main/java/se/inera/intyg/cts/infrastructure/service/SendPasswordWithSMS.java
```diff
@@ -44,21 +44,4 @@ public void sendPassword(Termination termination) {
         updateTermination.passwordSent();
```
mww-sopra-steria, line 44: I know this is not part of the PR, but same as with the SendPackageNotification class, any changes to the domain should not be done within the infrastructure layer.

customer-terminate-service/app/src/main/resources/db/changelog/db.changelog-master.xml
```diff
@@ -116,4 +116,12 @@
     </addColumn>
   </changeSet>
 
+  <changeSet id="6" author="mh">
```
mww-sopra-steria, line 119: How will this impact any already existing terminations as it cannot be null? Add a default value like change@me.se or something?

mww-sopra-steria (changes requested):
Nice! A couple of comments to look into. I have also been thinking if the reminder logic should be moved into the domain... 

Main thought are that the domain package should include all the logic related to the business.

mww-sopra-steria (approved):

PR #18, opened 2022-08-17
INTYGFV-14571: Logic to update certain metadata in an existing termination.

mww-sopra-steria:


customer-terminate-service/app/src/main/java/se/inera/intyg/cts/application/service/TerminationService.java
```diff
@@ -50,6 +58,32 @@ public TerminationDTO create(CreateTerminationDTO createTerminationDTO) {
     return toDTO(createdTermination);
   }
 
+  public TerminationDTO update(UUID terminationId, TerminationDTO terminationDTO) {
+    if (!terminationId.equals(terminationDTO.terminationId())) {
```
IamFriberg, line 62: Feel a bit redundant to pass the same parameter 2 times to this method. Unless we can update all the fields in TerminationDTO we should not use it as a parameter. (Now we only update/use 4 of 10 fields/arguments provided.)

customer-terminate-service/app/src/main/java/se/inera/intyg/cts/infrastructure/persistence/JpaCertificateRepository.java
```diff
@@ -51,6 +51,23 @@ public List<Certificate> get(Termination termination) {
         .collect(Collectors.toList());
   }
 
+  @Override
+  public void remove(Termination termination) {
+    final var terminationEntity = terminationEntityRepository
+        .findByTerminationId(termination.terminationId().id());
+
+    if (terminationEntity.isEmpty()) {
+      return;
```
IamFriberg, line 60: Maybe we should throw a NotFoundException if we expect to find a termination to remove?

IamFriberg (changes requested):
Nice! Only have som smaller comments around the amount of field we pass to the update.

customer-terminate-service/app/src/main/java/se/inera/intyg/cts/infrastructure/persistence/JpaCertificateRepository.java
mww-sopra-steria: You have a good point. It is better to do as you suggest as it is correct that we do expect a termination to exist. 

Will fix!

customer-terminate-service/app/src/main/java/se/inera/intyg/cts/infrastructure/persistence/JpaCertificateTextRepository.java
```diff
@@ -52,4 +52,24 @@ public List<CertificateText> get(Termination termination) {
         .map(CertificateTextEntityMapper::toDomain)
         .collect(Collectors.toList());
   }
+
+  @Override
+  public void remove(Termination termination) {
+    final var terminationEntity = getByTerminationId(termination);
+
+    if (terminationEntity.isEmpty()) {
+      return;
```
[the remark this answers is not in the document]
mww-sopra-steria: You have a good point. It is better to do as you suggest as it is correct that we do expect a termination to exist. 

Will fix!

customer-terminate-service/domain/src/main/java/se/inera/intyg/cts/domain/service/UpdateTerminationImpl.java
```diff
@@ -0,0 +1,41 @@
...
+
+  @Override
+  public Termination update(Termination termination, HSAId hsaId, PersonId personId,
+      EmailAddress emailAddress, PhoneNumber phoneNumber) {
+
+    termination.update(hsaId, personId, emailAddress, phoneNumber);
+
+    final var storeTermination = terminationRepository.store(termination);
+    if (storeTermination.status() == TerminationStatus.CREATED) {
+      certificateRepository.remove(storeTermination);
```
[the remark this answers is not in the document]
mww-sopra-steria: We don't remove the termination, we remove any certificates and certificate texts that are related to the termination. 

It is an important behaviour for the service and there are unit tests that verifies it

customer-terminate-service/app/src/main/java/se/inera/intyg/cts/application/service/TerminationService.java
mww-sopra-steria: I didn't want to put that logic in the controller, instead this app service verifies the data received through the api and then calls the domain service with the fields to update. If in tomorrow we allow another field to be updated, we don't have to change the api.

IamFriberg: With this solution we allow the caller to provide arguments that are not used but can be provided. This make the API unsafe as we can not trust the service to updated all fields we provide. Another solution would be to use a updateTerminationDTO what only contains the 3 fields that will be used and give us the option to add optional fields in the future that will not affect the API.

IamFriberg (changes requested):

customer-terminate-service/domain/src/main/java/se/inera/intyg/cts/domain/service/UpdateTerminationImpl.java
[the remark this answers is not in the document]
IamFriberg: Should we add some description to why we do this? My picture is that this is a part of a chain of events that does not realy make sense unless you know the whole chain and where it fits. :)

customer-terminate-service/app/src/main/java/se/inera/intyg/cts/infrastructure/persistence/JpaCertificateRepository.java
mww-sopra-steria: Fixed!

customer-terminate-service/app/src/main/java/se/inera/intyg/cts/infrastructure/persistence/JpaCertificateTextRepository.java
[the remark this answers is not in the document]
mww-sopra-steria: Fixed!

customer-terminate-service/domain/src/main/java/se/inera/intyg/cts/domain/service/UpdateTerminationImpl.java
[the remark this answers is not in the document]
mww-sopra-steria: You have a good point. I will some javadoc that points out this important behaviour.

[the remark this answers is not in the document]
mww-sopra-steria: Fixed!

customer-terminate-service/app/src/main/java/se/inera/intyg/cts/application/api/TerminationController.java
```diff
@@ -37,6 +37,16 @@ TerminationDTO create(@RequestBody CreateTerminationDTO request) {
     }
   }
 
+  @PostMapping("/{terminationId}")
+  TerminationDTO update(@PathVariable UUID terminationId,
```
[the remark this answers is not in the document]
mww-sopra-steria: It makes sense, and I think is expected, that the apis url specifies which termination to update. That it also exists in the TerminationDTO is a bit redundant, but is validated in the TerminationService that there is no missmatch.

IamFriberg, line 41: With this solution we allow the caller to provide arguments that are not used but can be provided. This make the API unsafe as we can not trust the service to updated all fields we provide. Another solution would be to use a updateTerminationDTO what only contains the 3 fields that will be used and give us the option to add optional fields in the future that will not affect the API.

IamFriberg (changes requested):

customer-terminate-service/app/src/main/java/se/inera/intyg/cts/application/api/TerminationController.java
mww-sopra-steria: OK! I do get your point and I agree. Will make the change and communicate with @mikeyOne to make necessary changes in Intygsadmin.

mww-sopra-steria: Fixed!

IamFriberg (approved):
Nice!

PR #19, opened 2022-08-23
INTYGFV-14568: Introduced timestamps for modified, exported, notification and reminder. Use notification timestamp when decided if a reminder should be sent.

mww-sopra-steria:


customer-terminate-service/domain/src/test/java/se/inera/intyg/cts/domain/model/TerminationBuilderTest.java
```diff
@@ -88,12 +103,24 @@ void shallNotExceptExistingTerminationWithoutCreated() {
       assertEquals("Missing Created", exception.getMessage());
     }
 
+    @Test
+    void shallNotExceptExistingTerminationWithoutModified() {
```
mikeyOne, line 107: shallNotAccept... ? Also the above and below method :)

mikeyOne (approved):

customer-terminate-service/domain/src/test/java/se/inera/intyg/cts/domain/model/TerminationBuilderTest.java
mww-sopra-steria: Fixed!

PR #27, opened 2023-09-03
INTYGFV-15836: Added new service to act as proxy, named intyg-proxy-service, when integrating with external services. This commit contains a skeleton to integrate with PUService, which will be the first integration to act as a proxy for.

mww-sopra-steria:


mikeyOne (approved):
Looks good as a starting point!

PR #29, opened 2023-09-07
INTYGFV-15838: Integration with PU.

mww-sopra-steria:


mhornfeldt (approved):
Nice! The only thing I thought about was if we wanted to name it FAKE_PU_PROFILE instead of FAKEPU_PROFILE. But both work 👍

PR #32, opened 2023-09-19
INTYGFV-15841: Added logback and MDC management.

ChrilleSigerhed:


intyg-proxy-service/devops/dev/config/logback-spring.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<configuration>
+  <property name="APP_NAME" value="${APP_NAME:-intyg-proxy-}"/>
```
mww-sopra-steria, line 3: Remove APP_NAME as it is not used for intyg-proxy-service.

intyg-proxy-service/build.gradle
```diff
@@ -11,6 +11,7 @@ ext {
     jaxsWsVersion = "4.0.1"
     jakartaWsVersion = "4.0.0"
     springbootMavenBomVersion = '3.1.2'
+    guavaVersion = '30.1.1-jre'
```
mww-sopra-steria, line 14: Can you see if we can go for an updated guavaVersion? This version is 2 years old and contains vulnerabilities
https://mvnrepository.com/artifact/com.google.guava/guava/30.1.1-jre

mww-sopra-steria (changes requested):
Sorry to keep you waiting for a review... Looks good but take a look at the comments. Also, is it possible to add some unit tests?

intyg-proxy-service/devops/dev/config/logback-spring.xml
ChrilleSigerhed: Removed

intyg-proxy-service/build.gradle
ChrilleSigerhed: Bumped to 32.1.2-jre. Was commonly used and has no vulnerabilities.

https://mvnrepository.com/artifact/com.google.guava/guava

ChrilleSigerhed:
I've tried implementing tests for the MDC filter, but im not able to make a correct assertion. Im guessing its related to that the MDC.clear() is called in finally {}. Could also be related to MDC threading?

Also tried adding tests for the MarkerFilter, but not able to work around this method:
```java
@Override
  public void start() {
    if (!this.markersToMatch.isEmpty()) {
      super.start();
    } else {
      addError("!!! no marker yet !!!");
    }
  }
```

mww-sopra-steria (approved):
Nice!

PR #36, opened 2023-10-31
INTYGFV-16032: Added getHealthCareUnitMembers to intyg proxy

mhornfeldt:


mww-sopra-steria (approved):
Looks good!

Could you do two minor refactorings before merging? This is to be more consistent with the pattern in minaintyg.

1) in integration-api rename the package dto -> model
2) in integration-api remove the service package and let the interfaces sit directly "in" Organization.

PR #37, opened 2023-11-01
INTYGFV-16029: Added configuration for hsa webservices. Refactored common integration code to integration-common module. Restructured integration properties.

mww-sopra-steria:


mhornfeldt (approved):
Nice that you did a factory

PR #40, opened 2023-11-02
INTYGFV-16027: Added tk implementation of get health care unit

mhornfeldt:


mww-sopra-steria (approved):
Hopefully we can get rid of the need for TypeConverterHelper once we get the code generation to work as we want. Everything looks good, just a small thing to add before you merge. But I will approve it.

PR #44, opened 2023-11-24
INTYGFV-16138: Added get unit integration

mhornfeldt:


intyg-proxy-service/integration-hsa/src/main/java/se/inera/intyg/intygproxyservice/integration/organization/client/converter/GetUnitResponseTypeConverter.java
```diff
@@ -0,0 +1,40 @@
...
+import se.riv.infrastructure.directory.organization.getunitresponder.v4.GetUnitResponseType;
+
+@Service
+@RequiredArgsConstructor
+public class GetUnitResponseTypeConverter {
+
+  private final UnitTypeConverter unitTypeConverter;
+
+  public Unit convert(GetUnitResponseType type) {
+    if (type == null || type.getUnit() == null) {
```
ChrilleSigerhed, line 34: If the unit is not found, might be a good idea to log this event, to keep existing functionality. Maybe something like:

```java
log.warn("No unit was found with hsaId '{}', returning empty unit", getUnitRequestDTO.getHsaId());
```

intyg-proxy-service/integration-hsa/src/main/java/se/inera/intyg/intygproxyservice/integration/organization/client/converter/UnitTypeConverter.java
```diff
@@ -0,0 +1,69 @@
...
+        .mail(type.getMail())
+        .unitStartDate(toLocalDate(type.getUnitStartDate()))
+        .unitEndDate(toLocalDate(type.getUnitEndDate()))
+        .postalCode(type.getPostalCode())
+        .postalAddress(addressTypeConverter.convertV3(type.getPostalAddress()))
+        .careType(type.getCareType())
+        .businessType(type.getBusinessType())
+        .management(type.getManagement())
+        .geographicalCoordinatesRt90(
+            geoCoordRt90TypeConverter.convert(type.getGeographicalCoordinatesRt90())
```
ChrilleSigerhed, line 57: In existing implementation a null-check is added before getGeographicalCoordinatesRt90 is set
```java
if (unitType.getGeographicalCoordinatesRt90() != null) {
    unit.setGeographicalCoordinatesRt90(toRt90(unitType.getGeographicalCoordinatesRt90()));
}
```
Should we add a similar check in the ```GeoCoordRt90TypeConverter```, to avoid NullPointerException?

intyg-proxy-service/integration-hsa/src/main/java/se/inera/intyg/intygproxyservice/integration/organization/client/converter/UnitTypeConverter.java
```diff
@@ -0,0 +1,69 @@
...
+        )
+        .geographicalCoordinatesSweref99(
```
ChrilleSigerhed, line 59: In existing implementation a null-check is added before geographicalCoordinatesSweref99 is set
```java
     if (unitType.getGeographicalCoordinatesSWEREF99() != null) {
            unit.setGeographicalCoordinatesSweref99(toSweref99(unitType.getGeographicalCoordinatesSWEREF99()));
        }
```
Should we add a similar check in the ```GeoCoordSweref99TypeConverter```, to avoid NullPointerException?

intyg-proxy-service/integration-hsa/src/main/java/se/inera/intyg/intygproxyservice/integration/organization/client/HsaOrganizationClient.java
```diff
@@ -71,6 +89,13 @@ public HealthCareUnitMembers getHealthCareUnitMembers(
     return getHealthCareUnitMembersResponseTypeConverter.convert(type);
   }
 
+  private static GetUnitType getUnitParameters(String hsaId) {
+    final var parameters = new GetUnitType();
+    parameters.setUnitHsaId(hsaId);
+    parameters.setIncludeFeignedObject(false);
```
ChrilleSigerhed, line 95: We also need to set a Profile right? 
These are our alternatives:
```java
    @XmlEnumValue("basic")
    BASIC("basic"),
    @XmlEnumValue("extended")
    EXTENDED("extended"),
    @XmlEnumValue("all")
    ALL("all");
```
I suggest we set it to ```basic```

ChrilleSigerhed (changes requested):
Great job, added some comments to consider.

Also, in ```UnitService``` we have a log saying:
```java
log.info(String.format("Unit with hsaId: '%s' was retrieved", request.getHsaId()));
```
This may be false, since if the unit was not found, we return an empty unit. Consider moving this log closer to the source so that we dont log the wrong information.

intyg-proxy-service/integration-hsa/src/main/java/se/inera/intyg/intygproxyservice/integration/organization/client/converter/GetUnitResponseTypeConverter.java
mhornfeldt: I haven't added any logging for the other implementations when receiving null. Is this function the only one that has it or should it maybe not be placed in IPS?

intyg-proxy-service/integration-hsa/src/main/java/se/inera/intyg/intygproxyservice/integration/organization/client/HsaOrganizationClient.java
mhornfeldt: I tried adding profile before but from what I can see it doesn't have a setter

intyg-proxy-service/integration-hsa/src/main/java/se/inera/intyg/intygproxyservice/integration/organization/client/converter/GetUnitResponseTypeConverter.java
mhornfeldt: Separated jira is created

ChrilleSigerhed (approved):
Awsome, great work!

PR #45, opened 2023-11-24
INTYGFV-16137 & INTYGFV-16136: Implemented get credential information

mhornfeldt:


intyg-proxy-service/integration-hsa/src/main/java/se/inera/intyg/intygproxyservice/integration/authorization/client/converter/CredentialInformationTypeConverter.java
```diff
@@ -0,0 +1,66 @@
...
+        .hsaSystemRole(
+            type.getHsaSystemRole().stream()
+                .map(hsaSystemRoleTypeConverter::convert)
+                .toList()
+        )
+        .build();
+  }
+
+  private String convertPersonalIdentity(IIType type) {
+    return type == null ? "" : type.getExtension();
```
ChrilleSigerhed, line 64: In current implementation this value is only set if IIType != null. Consider instead of returning an empty string, to set it to null.

```java
return type != null ? type.getExtension() : null;
```

ChrilleSigerhed (approved):
Awsome!

PR #53, opened 2023-12-08
INTYGFV-16174 & INTYGFV-16176: Implemented integration for get credentials for person

mhornfeldt:


intyg-proxy-service/integration-hsa/src/main/java/se/inera/intyg/intygproxyservice/integration/authorization/client/converter/GetCredentialsForPersonResponseTypeConverter.java
```diff
@@ -0,0 +1,60 @@
+package se.inera.intyg.intygproxyservice.integration.authorization.client.converter;
+
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Service;
+import se.inera.intyg.intygproxyservice.integration.api.authorization.model.CredentialsForPerson;
+import se.riv.infrastructure.directory.authorizationmanagement.gethospcredentialsforpersonresponder.v1.GetHospCredentialsForPersonResponseType;
+import se.riv.infrastructure.directory.authorizationmanagement.v2.IIType;
+
+@Service
```
ChrilleSigerhed, line 9: I dont think it really matters, but the other components are annotated with Component instead. Maybe we should do the same for all converters?

ChrilleSigerhed (approved):
Nice, great that you added the null check on type for both responses in the converters.

PR #91, opened 2024-02-04
INTYGFV-16343: Logic to handle requirements relating to protected persons when creating and retrieving certificates.

mww-sopra-steria:
Added some testdata and also fixed a mistake with using Alve instead of Atlas (deceased).


ChrilleSigerhed (approved):
Well done!

PR #93, opened 2024-02-06
INTYGFV-16356: Refactor getters in domain to use fluent accessors e.g name() instead of getName().

mww-sopra-steria:


mikeyOne (approved):
looks great

ChrilleSigerhed (approved):
Nicely done, it feels very satisfying that our getters are handled the same way regardless of whether the different objects being fetched are records or regular classes with value annotation in our domain.

PR #94, opened 2024-02-07
INTYGFV-16359: Added support to get supported certificate types in testability.

ChrilleSigerhed:


mww-sopra-steria (approved):
Nicely done! Like the way you build the response and merge certificate types and their versions.

PR #95, opened 2024-02-08
INTYGFV-16353: Expanded the domain model to include specification of certificate content and logic to handle update certificates in the domain

mww-sopra-steria:


ChrilleSigerhed (approved):
Well done, impressive structure of the domain logic. Very clean and easy to follow and understand.

PR #97, opened 2024-02-09
INTYGFV-16290: Added endpoint to handle update of certificate data. Added converters to update CertificateDataElements information. Added Resourcelinks to CertificateDTO.

ChrilleSigerhed:


ChrilleSigerhed:
image

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/certificate/dto/UpdateCertificateRequest.java
```diff
@@ -0,0 +1,23 @@
...
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import se.inera.intyg.certificateservice.application.common.dto.PatientDTO;
+import se.inera.intyg.certificateservice.application.common.dto.UnitDTO;
+import se.inera.intyg.certificateservice.application.common.dto.UserDTO;
+
+@Data
```
mww-sopra-steria, line 11: Great if you could change this to @Value @Builder

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/certificate/dto/config/CertificateDataConfigTypes.java
```diff
@@ -0,0 +1,8 @@
+package se.inera.intyg.certificateservice.application.certificate.dto.config;
+
+public enum CertificateDataConfigTypes {
```
mww-sopra-steria, line 3: What you think of us removing the types that we don't support?

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/certificate/dto/validation/CertificateDataValidationType.java
```diff
@@ -0,0 +1,7 @@
+package se.inera.intyg.certificateservice.application.certificate.dto.validation;
+
+public enum CertificateDataValidationType {
```
mww-sopra-steria, line 3: What you think to just include the types CS support?

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/certificate/service/CertificateDataConverter.java
```diff
@@ -0,0 +1,114 @@
...
+
+@Component
+@RequiredArgsConstructor
+public class CertificateDataConverter {
+
+  private final CertificateDataValueConverter certificateDataValueConverter;
+  private final CertificateDataConfigConverter certificateDataConfigConverter;
+  private final CertificateDataValidationConverter certificateDataValidationConverter;
+
+  public Map<String, CertificateDataElement> convert(CertificateModel certificateModel,
```
mww-sopra-steria, line 30: Good job! It's a bit complicated code, but it does the work and you have build a great unit test suite to verify that. 

What I would do is try to flatten and convert at the same time. I think that would simplify the code and make it more readable. It won't be too simple though, because we need to use recursive functions.

I don't expect you to change it and we can go ahead with this. But if you like to take a look at it, here is an article that flattens a hierarchical list structure: https://stackoverflow.com/questions/52670243/recursively-flatten-a-list-with-streams

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/certificate/service/UpdateCertificateService.java
```diff
@@ -0,0 +1,52 @@
...
+        updateCertificateRequest.getPatient(),
+        updateCertificateRequest.getUser(),
+        updateCertificateRequest.getUnit(),
+        updateCertificateRequest.getCareUnit(),
+        updateCertificateRequest.getCareProvider()
+    );
+
+    final var elementDataList = updateCertificateRequest.getCertificate().getData()
+        .entrySet()
+        .stream()
```
mww-sopra-steria, line 37: A thought is to filter any data that don't need to be converted.. for example filter away any data that is of category. Instead of filter nonNull afterwards.

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificate/model/Certificate.java
```diff
@@ -20,7 +21,8 @@ public class Certificate {
   private final CertificateModel certificateModel;
   private final LocalDateTime created;
   private CertificateMetaData certificateMetaData;
-  private List<ElementData> elementData;
+  @Builder.Default
+  private List<ElementData> elementData = new ArrayList<>();
```
mww-sopra-steria, line 25: Thinking that the default value should be an immutable list. So there is no chance that "someone" adds things to it. It should only be updated with updateData(...) and when the Certificate is being built.

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificatemodel/model/ElementSpecification.java
```diff
@@ -10,9 +11,12 @@ public class ElementSpecification {
 
   ElementId id;
   ElementConfiguration configuration;
-  List<ElementRule> rules;
-  List<ElementValidation> validations;
-  List<ElementSpecification> children;
+  @Builder.Default
+  List<ElementRule> rules = new ArrayList<>();
```
mww-sopra-steria, line 15: Same for the Certificate. Give it default values of immutable lists.

certificate-service/integration-test/src/test/java/se/inera/intyg/certificateservice/integrationtest/fk7211/FK7211ActiveIT.java
```diff
@@ -513,4 +519,216 @@ void shallReturnCertificateIfPatientIsProtectedPersonAndUserIsDoctor() {
...
+      );
+
+      assertEquals(expectedDate,
+          getValueFromData(response.getBody(), CertificateDataValueType.DATE, questionId)
+      );
+    }
+
+    @Test
+    @DisplayName("FK7211 - Vårdadmin - Om intyget är utfärdat på en patient som har skyddade personuppgifter skall felkod 403 (FORBIDDEN) returneras")
+    void shallNotUpdateDataIfUserIsCareAdminAndPatientIsProtectedPerson() {
```
mww-sopra-steria, line 560: We don't need to do any actual updates... just call the api to update and it will give the error.

mww-sopra-steria (changes requested):
Awesome work! Most of it looks great. A few comments to consider, but not necessarily change anything.

I would like to discuss the integration tests separately. Maybe we could do that first thing on Monday.

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/certificate/dto/config/CertificateDataConfigTypes.java
ChrilleSigerhed: I agree!

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/certificate/service/UpdateCertificateService.java
ChrilleSigerhed: I like it, done!

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/certificate/service/UpdateCertificateService.java
```diff
@@ -0,0 +1,61 @@
...
+
+    final var updatedCertificate = updateCertificateDomainService.update(
+        new CertificateId(certificateId),
+        elementDataList,
+        actionEvaluation
+    );
+
+    return UpdateCertificateResponse.builder()
+        .certificate(certificateConverter.convert(updatedCertificate))
+        .build();
```
mww-sopra-steria, line 53: Missed it before, but you should add some resource links as well. Should work by just adding the same code as in GetCertificateService

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/certificate/service/CertificateDataConverter.java
ChrilleSigerhed: I made a try at flattening out the code, and I think it improved the readability of it. Lets discuss improvements further on monday!

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/certificate/service/UpdateCertificateService.java
mww-sopra-steria: I think we made a misstake with converting the ResourceLinks... the links attribute sits inside the CertificateDTO and not on the same level as the "certificate".

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificate/model/Certificate.java
ChrilleSigerhed: If I understand correct, we want to remove the Builder.Default annotation.

In some scenarios, like when we create or get a certificate, the element data list might not have been initialized and is null. If we dont want to use the Builder.Default-annotation, we need to introduce a null check, something like this:

CertificateConverter:
```java
      .data(
            certificateDataConverter.convert(
                certificate.certificateModel(),
                certificate.elementData() != null ? certificate.elementData() : Collections.emptyList()
            )
        )
```

Is that a more appropriate way to handle the chance of ElementData beeing null?

Alternatively, we could introduce a get-function in Certificate to get ElementData:

```java
  public List<ElementData> getElementData() {
    if (this.elementData == null) {
      return Collections.emptyList();
    }
    return this.elementData;
  }
```

mww-sopra-steria: No, I didn't mean to remove the default-annotation. Instead the default value should be an immutable list. 

If someone updates the elementData, it would only be through the updateData(...) and that replaced the current list.

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/certificate/service/UpdateCertificateService.java
mww-sopra-steria: I been looking into this a little deeper and we have an issue actually.. We will need do make some changes to Webcert (and Common) in order to make this work correctly.

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificate/model/Certificate.java
ChrilleSigerhed: Ah I see, great catch!

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/certificate/service/UpdateCertificateService.java
ChrilleSigerhed: I see, I will abort the attempt to update links and maybe we can sync the required changes tomorrow morning.

mww-sopra-steria (approved):
Looks great! Well done!

PR #106, opened 2024-02-18
INTYGFV-16389: Implemented endpoint to query certificates on units. Refactored persistence layer to use jpa specifications for querying. Organized the domain layers models and services into different packages.

mww-sopra-steria:


ChrilleSigerhed (approved):
Impressive work, I really like the usage of Specifications! 

Well done!

PR #107, opened 2024-02-19
INTYGFV-16290: Added ConcurrentModificationException thats thrown if revision dont match in Certificate.

ChrilleSigerhed:


mww-sopra-steria (approved):
Looks great! And great that you use http-status CONFLICT!

PR #109, opened 2024-02-21
INTYGFV-16397: Temporarly fix to include resource link SIGN_CERTIFICATE.

ChrilleSigerhed:


mww-sopra-steria (approved):
Good pragmatic solution for something we will change when implementing "signing"-functionality!

PR #111, opened 2024-02-25
INTYGFV-16401: Introduced CertificateEvent to domain and a way to publish events to subscribers. Implemented a CertificateEventLogService that subscribes to domain events and log them conforming to ECS. Publish domain events for all the domain services in certificate.

mww-sopra-steria:


ChrilleSigerhed (approved):
Awsome work, well done! Great that we could utilize MDC to manipulate the fields of the ecs-logging!

Approved!

PR #118, opened 2024-03-11
INTYGFV-16412: Testing to create generic PDF using PdfBox

mikaelaromild1:


mww-sopra-steria:
These PRs should not be merged as they were part of a PoC. Will close them.

PR #122, opened 2024-03-18
INTYGFV-16440: Logic to support signing without signature for private doctors

mww-sopra-steria:


ChrilleSigerhed (approved):
Very impressive work, great job!

PR #125, opened 2024-03-21
INTYGFV-16465: Refactored signing and fixed issue that the xml wasn't correctly generated (missing signing and sending datetime)

mww-sopra-steria:


mhornfeldt (approved):
Nice!

PR #128, opened 2024-03-25
INTYGFV-16461: Implementation of endpoint to send certificates

mww-sopra-steria:


ChrilleSigerhed (approved):
Well done, and nice solution to the hibernate issue with StaffEntity.

PR #134, opened 2024-04-02
INTYGFV-16481: Added midwife role to action rules

mhornfeldt:


certificate-service/integration-test/src/test/java/se/inera/intyg/certificateservice/integrationtest/fk7211/FK7211ActiveIT.java
```diff
@@ -2664,6 +2665,28 @@ void shallReturn403IfUnitIsCareUnitAndNotOnCareUnit() {
...
     }
 
+    @Test
+    @DisplayName("FK7211 - Om användaren har rollen barnmorska ska intyget gå att signeras")
+    void shallSuccessfullySignIfRoleIsMidwife() {
+      final var testCertificates = testabilityApi.addCertificates(
+          defaultTestablilityCertificateRequest(FK7211, VERSION)
+      );
+
+      final var response = api.signCertificate(
```
ChrilleSigerhed, line 2675: Since the send-action also is affected by the role check, should we add a IT that verifies midwife can send certificate aswell?

ChrilleSigerhed (changes requested):
Added a comment regarding IT-test to consider.

ChrilleSigerhed (approved):
Well done, works like a charm! 8)

PR #143, opened 2024-04-10
INTYGFV-16409: Schematron validation for FK7211 & FK7443.

ChrilleSigerhed:


certificate-service/clinicalprocess-certificate-v4/src/test/resources/itfp.v1.sch
```diff
@@ -0,0 +1,102 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mww-sopra-steria, line 1: The test classes should not use its own schematron. It should use the actual schematrons for the certificates. Otherwise we might have a schematron that works in the test class but not in production.

certificate-service/clinicalprocess-certificate-v4/src/test/resources/igrav.v1.sch
```diff
@@ -0,0 +1,66 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mww-sopra-steria, line 1: The test classes should not use its own schematron. It should use the actual schematrons for the certificates. Otherwise we might have a schematron that works in the test class but not in production.

mww-sopra-steria (changes requested):
See two comments about schematron files and tests.

One thing that I would like us to consider/discuss is making the schema-validation and schematron-validation part of the XmlGenerator (should of course be separate implementation-classes as in this PR).

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificate/service/XmlSchemaValidator.java
```diff
@@ -0,0 +1,10 @@
+package se.inera.intyg.certificateservice.domain.certificate.service;
+
+import se.inera.intyg.certificateservice.domain.certificate.model.CertificateId;
+import se.inera.intyg.certificateservice.domain.certificate.model.Xml;
+
+public interface XmlSchemaValidator {
```
mww-sopra-steria, line 6: Move this interface to clinicalprocess gradle module as its not necessary to expose it as part of the domain.

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificate/service/XmlSchematronValidator.java
```diff
@@ -0,0 +1,11 @@
+package se.inera.intyg.certificateservice.domain.certificate.service;
+
+import se.inera.intyg.certificateservice.domain.certificate.model.CertificateId;
+import se.inera.intyg.certificateservice.domain.certificate.model.Xml;
+import se.inera.intyg.certificateservice.domain.certificatemodel.model.SchematronPath;
+
+public interface XmlSchematronValidator {
```
mww-sopra-steria, line 7: Move this interface to clinicalprocess gradle module as its not necessary to expose it as part of the domain.

mww-sopra-steria (approved):
Nice work! 

Suggest that we move the interfaces out from the domain before merging. See comments.

PR #148, opened 2024-04-12
INTYGFV-16515: Fix to correct database truncation for revoke

mhornfeldt:


mww-sopra-steria (approved):
Approved! I'm thinking that we should refactor the revoke part to a separate table, now when we have a large field like revoked_message. This is something we can discuss on the Sprint planning, but this PR is a good fix to make sure the revoke-functionality works as expected!

PR #149, opened 2024-04-14
INTYGFV-16489: Added intygsgivare to xml format.

mikeyOne:


mww-sopra-steria (approved):
Looks good! Its special handing of "intygsutgivare" and the way you implemented it is as simple as possible. 

We will refactor the "svar"-converters when we implement additional certificate types.

PR #151, opened 2024-04-16
INTYGFV-16524: Added label to question period in CertificateModelFactoryFK7443.

ChrilleSigerhed:


certificate-service/app/src/test/java/se/inera/intyg/certificateservice/application/certificate/service/converter/CertificateDataCheckboxDateRangeListConfigConverterTest.java
```diff
@@ -62,6 +63,7 @@ void shouldReturnConvertedConfig() {
                     .hideWorkingHours(true)
                     .previousDateRangeText("TEST TEST")
                     .name("NAME")
+                    .label("LABEL")
                     .dateRanges(
```
mhornfeldt, line 67: We can look into the name. Do we want to change it to more descriptive that it's a level under "name" or have a list of "subtexts"

mhornfeldt: We did a solution in Mina intyg where we had a list of texts which had an id to show what text it was

ChrilleSigerhed: Implemented new record called SubText that holds an map of type and text. Maybe this is a better more generic way to handle different types of texts in our model?

Not sure about the name SubText tho, what do you think?

mhornfeldt: I think we should call it texts/text or similar so it's more general. That means that we could include texts that are not subtexts if we need to in the future (info, alerts etc.). You can then describe that it's a sub text or similar with the id. What do you think?

ChrilleSigerhed: Changed the field to texts and renamed the record to QuestionTexts to be more generic.

mww-sopra-steria (changes requested):
Suggest that we don't introduce QuestionText as of yet. Cant see that it is a generic enough concept that adds value instead of just introducing  attribute for label.

We could consider it again at a later stage if the need present itself.

mww-sopra-steria (approved):

PR #153, opened 2024-04-17
INTYGFV-16519: Added earliest limit to date range

mikaelaromild1:


certificate-service/domain/src/test/java/se/inera/intyg/certificateservice/domain/validation/model/ElementValidationDateRangeListTest.java
```diff
@@ -930,6 +1024,99 @@ void shouldReturnNoValidationErrorsIfCorrectPeriodsWithNoOverlap() {
 
       assertEquals(Collections.emptyList(), actualResult);
     }
+
+    @Nested
+    class BeforeMinDate {
```
ChrilleSigerhed, line 1029: Please also add some tests to verify behaviour when min is not set.

For example, take a look at ```ElementValidationTextTest``` that handles both mandatory and limit tests.

ChrilleSigerhed (changes requested):
Looks great, just need to add some tests to verify min validation is not active if not set.

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/fk7443/CertificateModelFactoryFK7443.java
```diff
@@ -8,6 +8,7 @@
 import static se.inera.intyg.certificateservice.domain.certificatemodel.model.WorkCapacityType.TRE_FJARDEDELAR;
 
 import java.time.LocalDateTime;
+import java.time.Period;
```
mhornfeldt, line 11: For date validation we had min set as type TemporalAmount. Do we want to have the same here?

ChrilleSigerhed (approved):
Works great, good job! :)

PR #154, opened 2024-04-19
INTYGFV-16522: Refactor pdf generator and add generator for Fk7443

mhornfeldt:


certificate-service/pdfbox-generator/src/main/java/se/inera/intyg/certificateservice/pdfboxgenerator/FK7211PdfGenerator.java
```diff
@@ -35,186 +22,56 @@ public class FK7211PdfGenerator {
...
-  public FK7211PdfGenerator(Certificate certificate) {
-    final var template = certificate.certificateModel().pdfTemplatePath();
-    try (final var inputStream = getClass().getClassLoader().getResourceAsStream(template)) {
-      if (inputStream == null) {
-        throw new IllegalArgumentException("Template not found: " + template);
-      }
-      pdDocument = Loader.loadPDF(inputStream.readAllBytes());
-      final var documentCatalog = pdDocument.getDocumentCatalog();
-      acroForm = documentCatalog.getAcroForm();
+  public static final ElementId QUESTION_BERAKNAT_NEDKOMSTDATUM_ID = new ElementId("1");
```
mhornfeldt, line 25: Feels a bit unnatural to copy paste in the question ids here. But were in a different package and I don't think the right way to go is to add a dependency to where these ids exists. Also, I was thinking that we could programatically match the questions filled in to the fields in the pdfs. But then we would have to add the pdf field id's to the certificate model, which might also be something we don't want to do.

certificate-service/pdfbox-generator/src/main/java/se/inera/intyg/certificateservice/pdfboxgenerator/FK7211PdfFillService.java
```diff
@@ -0,0 +1,72 @@
...
+import java.io.IOException;
+import org.apache.pdfbox.pdmodel.interactive.form.PDAcroForm;
+import se.inera.intyg.certificateservice.domain.certificate.model.Certificate;
+import se.inera.intyg.certificateservice.domain.certificatemodel.model.CertificateType;
+import se.inera.intyg.certificateservice.domain.certificatemodel.model.ElementId;
+import se.inera.intyg.certificateservice.pdfboxgenerator.toolkits.PdfGeneratorValueToolkit;
+import se.inera.intyg.certificateservice.pdfboxgenerator.value.PdfDateValueGenerator;
+
+
+public class FK7211PdfFillService implements CertificateTypePdfFillService {
```
mww-sopra-steria, line 12: Move specific certificate type code to a sub package.

certificate-service/pdfbox-generator/src/test/resources/fk7443_v1.pdf
mww-sopra-steria, line 1: Remove pdf-files used for test. Instead add test-dependency to app-module and use the actual files. This minimises the risk of using the wrong file in test/production.

mww-sopra-steria (changes requested):
Nice job on refactoring the code and increase reusability.

A few general comments to consider:
1) Look at the package structure and class names. Now it toolkit, helper, util. It doesn't say much what it is. See if you can organise it in more meaningful names. Think the comparison to legacy app using controller, service, repository etc compared to certificate, patient, unit etc. 
2) Not asking you to rewrite the code, but I think reusability increases when you try to design with pure-functions. No sideeffects. Now the interfaces takes acroform and modify it in many places. You could have changed it to gather necessary information and then have generic code filling in the pdf. Testability is also increased by doing that.
3) It is OK to use Spring in the pdf-module. Then dependency injection can be used instead of managing it within the classes itself.

mhornfeldt:
> Nice job on refactoring the code and increase reusability.
> 
> A few general comments to consider:
> 
> 1. Look at the package structure and class names. Now it toolkit, helper, util. It doesn't say much what it is. See if you can organise it in more meaningful names. Think the comparison to legacy app using controller, service, repository etc compared to certificate, patient, unit etc.
> 2. Not asking you to rewrite the code, but I think reusability increases when you try to design with pure-functions. No sideeffects. Now the interfaces takes acroform and modify it in many places. You could have changed it to gather necessary information and then have generic code filling in the pdf. Testability is also increased by doing that.
> 3. It is OK to use Spring in the pdf-module. Then dependency injection can be used instead of managing it within the classes itself.

Fixed

mww-sopra-steria (approved):
Nice solution by introducing the "PdfField"! You have opened up new possibilities by separating the mapping to the model and the pdfbox-integration. It is also easy to read and understand the code. Well done!

Consider moving the packages "text" and "value" to "pdf" before merging.

PR #158, opened 2024-04-25
INTYGFV-16537: Extend the xml api to be able to fetch xml for drafts and stored xmls

mikaelaromild1:


mikaelaromild1:
@mww-sopra-steria In regards to the discussion we had:

Since the certificiate status is set to signed before the xml generator is called in the signing processes in **SignCertificateDomainService** and **SignCertificateWithoutSignatureDomainService**, I think it might be okay to just check if the certificate is a draft in **XmlGeneratorCertificateV4**. But I may be wrong here :)

ChrilleSigerhed (approved):
Well done!

mww-sopra-steria (approved):
Looks good!

PR #161, opened 2024-04-26
INTYGFV-16520: Added tags for added texts to PDF

mhornfeldt:


certificate-service/pdfbox-generator/src/main/java/se/inera/intyg/certificateservice/pdfboxgenerator/pdf/fk7443/FK7443PdfFillService.java
```diff
@@ -23,6 +23,8 @@ public class FK7443PdfFillService implements CertificateTypePdfFillService {
   public static final String PATIENT_ID_FIELD_ID = "form1[0].#subform[0].flt_txtPersonNrBarn[0]";
   public static final String PERIOD_FIELD_ID_PREFIX = "form1[0].#subform[0]";
 
+  private int mcid = 100;
```
mww-sopra-steria, line 26: The service has to be stateless. This introduces state and this integer will continue to grow for each time getAvailableMcid is called. We have to find a solution which doesn't introduce state in the service.

mww-sopra-steria (changes requested):
Great job finding a way to adding these tags! See comment relating to "state". We need to find a different way to keep track on mcid instead of keeping state within a service.

certificate-service/pdfbox-generator/src/main/java/se/inera/intyg/certificateservice/pdfboxgenerator/pdf/fk7443/FK7443PdfFillService.java
mww-sopra-steria: Added a fix with AtomicInteger and removed state in the services.

mww-sopra-steria (approved):
Approved! Did a peer-review with @ChrilleSigerhed as well.

PR #162, opened 2024-04-30
INTYGFV-16546: Updated API to handle inacitve units, access scope and reference parameters

mikaelaromild1:


certificate-service/integration-test/src/test/java/se/inera/intyg/certificateservice/integrationtest/fk7211/FK7211ActiveIT.java
```diff
@@ -3481,4 +3483,143 @@ void shallReturn403IfCertificateIsAlreadyRevoked() {
...
+              .build(),
+          certificateId(testCertificates)
+      );
+
+      assertEquals(403, response.getStatusCode().value());
+    }
+  }
+
+  @Nested
+  @DisplayName("FK7211 - Utökad behörighet vid SVOD")
```
mww-sopra-steria, line 3559: Suggest we include "djupintegration" in the diplayname. Maybe something like: "FK7211 - Utökat behörighet vid djupintegration och SVOD (sjf=true)". Then "FK7211 - Utökad behörighet vid djupintegration" can also add something saying it is without SVOD.

mww-sopra-steria (changes requested):
Well done! Nice to see a very robust test suite on the access-rule!

A few comments to look at, mostly suggestions.

PR #163, opened 2024-05-02
INTYGFV-16548: Implementation of CertificateExternalTypeInfo endpoint.

ChrilleSigerhed:


mww-sopra-steria (changes requested):
1) Add GET endpoint to CertificateTypeInfoController.
2) Add "match" instead of overriding equals and hashcode.

certificate-service/domain/src/test/java/se/inera/intyg/certificateservice/domain/common/model/CodeTest.java
```diff
@@ -0,0 +1,29 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.junit.jupiter.api.Test;
+
+class CodeTest {
+
+  private static final String DISPLAY_NAME = "displayName";
+
+  @Test
+  void shallReturnTrueIfCodeSystemAndCodeMatch() {
```
mww-sopra-steria, line 13: Maybe add a test that they should match even if display name doesn't match.

mww-sopra-steria (approved):
Looks good! Consider add a test as suggested in comment before merging.

PR #164, opened 2024-05-03
INTYGFV-16557: Implementation of app and domain layer for replace certificates

mww-sopra-steria:
Excluded in this PR is the feature to continue on a replaced certificates draft. Will be handled in a separate PR as it requires a little analysis and design to decide how it can be handled without a "hack".


ChrilleSigerhed (commented):
Should we handle the required updates needed in CertificateConverter related to the relations-attribute in this PR?

ChrilleSigerhed (approved):
Awsome work with this PR, great job!

PR #169, opened 2024-05-07
INTYGFV-16525: Set to use nvd mirror for dependencycheck. Required gr…

mikeyOne:
…adle update. Performed general version update and some sonar cleaning.


mww-sopra-steria (approved):
Looks great! Well done!

PR #172, opened 2024-05-08
INTYGFV-16574: Implemented endpoint for citizen to get certificate

mikaelaromild1:


certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/citizen/service/converter/CertificateTextConverter.java
```diff
@@ -0,0 +1,36 @@
...
+            .toList())
+        .build();
+  }
+
+  private CertificateTextTypeDTO convertCertificateTextType(
+      CertificateTextType certificateTextType) {
+    if (Objects.requireNonNull(certificateTextType) == CertificateTextType.PREAMBLE_TEXT) {
+      return CertificateTextTypeDTO.PREAMBLE_TEXT;
+    }
+    return null;
```
mww-sopra-steria, line 34: Don't we need to have a texttype? Should we instead throw exception here to fail fast, instead of an error happening on the client side?

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/citizen/service/converter/AvailableFunctionsFactory.java
```diff
@@ -0,0 +1,93 @@
...
+                        .type(InformationType.FILENAME)
+                        .text(getFileName(certificate))
+                        .build()
+                )
+            )
+            .build()
+    );
+  }
+
+  private boolean isSendActive(Certificate certificate) {
```
mww-sopra-steria, line 64: This logic should be part of the domain and not in the app-layer. Maybe the easiest thing to do now is to encapsulate the logic in the Certificate domain class. Name the method something like isSendActiveForCitizen. This is to ensure that we don't get domain logic in the app-layer. We can refactor it more later if necessary when the actual implementation of send (by citizen) is implemented.

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/fk7211/CertificateModelFactoryFK7211.java
```diff
@@ -115,6 +128,20 @@ public CertificateModel create() {
         )
         .pdfTemplatePath(PDF_FK_7211_PDF)
         .schematronPath(SCHEMATRON_PATH)
+        .summary(CertificateSummary.builder()
```
mww-sopra-steria, line 131: Instead of providing "data" instead provide an implementation of an interface that created a CertificateSummary.

For example:

interface CertificateSummaryProvider {
    CertificateSummary summaryOf(Certificate certificate);
}


class FK7211CertificateSummaryProvider implements CertificateSummaryProvider {
    ....
}

You can move the logic that currently sits in the "CertificateConverter.getCertificateSummaryValue(...)".

Reason for this is that the rules for a summary cannot easily be modelled the way it is now. Suggested solution gives more flexibility and possibility for any custom implementation for a summary.

mww-sopra-steria (changes requested):
Nicely done! Have some comments though relating to summary and logic to decide available functions. Don't think it will be big changes, but if you agree on the suggestions, it will move logic to the domain and the model.

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/citizen/service/PrintCitizenCertificateService.java
```diff
@@ -0,0 +1,35 @@
...
+
+@Service
+@RequiredArgsConstructor
+public class PrintCitizenCertificateService {
+
+  private final PrintCitizenCertificateDomainService printCitizenCertificateDomainService;
+
+  public PrintCitizenCertificateResponse get(PrintCitizenCertificateRequest request,
+      String certificateId) {
+
```
mww-sopra-steria, line 19: We should validate that the request is correct.

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificatemodel/model/FK7211CertificateSummaryProvider.java
```diff
@@ -0,0 +1,42 @@
+package se.inera.intyg.certificateservice.domain.certificatemodel.model;
+
+import java.time.format.DateTimeFormatter;
+import se.inera.intyg.certificateservice.domain.certificate.model.Certificate;
+import se.inera.intyg.certificateservice.domain.certificate.model.ElementValueDate;
+
+public class FK7211CertificateSummaryProvider implements CertificateSummaryProvider {
```
mww-sopra-steria, line 7: Move this implementation to infrastructure. It should sit together with the FK7211-factory class.

mww-sopra-steria (approved):
Well done! I've left a few comments that I would appreciate if we change before merge. But I approve it so you can merge as soon as it is done.

certificate-service/app/src/test/java/se/inera/intyg/certificateservice/application/citizen/service/PrintCitizenCertificateServiceTest.java
```diff
@@ -55,6 +56,21 @@ void shouldReturnPrintCitizenCertificateResponse() {
             ADDITIONAL_INFO_TEXT
         )
     ).thenReturn(PDF);
+  }
+
+  @Test
+  void shouldValidateCertificateId() {
```
mhornfeldt, line 62: We need test for validate person id also

mhornfeldt: Sorry my mistake just rename the test since you're validating both id and person id :)

PR #173, opened 2024-05-08
INTYGFV-16386: Added possibility to set enabled false in CertificateA…

mikeyOne:
…ctionCreate


mww-sopra-steria (changes requested):
I think we need to find a solution that isn't dependent on "state" within the class.

mww-sopra-steria (approved):
Looks great!

PR #174, opened 2024-05-13
INTYGFV-16436: Activate renew for fk7443 and don't include period when renewing

mww-sopra-steria:


ChrilleSigerhed (approved):
Nice, how smooth and easy it is to add renew functionality to a certificate now!

PR #176, opened 2024-05-15
INTYGFV-16576: Added healthCareProfessionalLicence & relations to XML.

ChrilleSigerhed:


mww-sopra-steria (approved):
Nice work! Looks great! Check the comment and add validation on user as discussed before merge.

PR #181, opened 2024-05-16
INTYGFV-16582: Implemented SEND_AFTER_SIGN CertificateAction.

ChrilleSigerhed:


certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/action/model/CertificateActionFactory.java
```diff
@@ -143,6 +143,14 @@ public static CertificateAction create(CertificateActionSpecification actionSpec
               )
           )
           .build();
+      case SEND_AFTER_SIGN -> CertificateActionSendAfterSign.builder()
+          .certificateActionSpecification(actionSpecification)
+          .actionRules(
+              List.of(
+                  new ActionRuleStatus(List.of(Status.SIGNED))
```
mhornfeldt, line 150: I'm wondering about the action rule status set to SIGN. For SIGN it is set to DRAFT. But maybe a certificate that is supposed to be signed and sent have both SIGN and SEND_AFTER_SIGN?

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/action/model/CertificateActionFactory.java
```diff
@@ -143,6 +143,14 @@ public static CertificateAction create(CertificateActionSpecification actionSpec
               )
           )
           .build();
+      case SEND_AFTER_SIGN -> CertificateActionSendAfterSign.builder()
+          .certificateActionSpecification(actionSpecification)
+          .actionRules(
```
mhornfeldt, line 148: Do we need to validate that it has not already been sent?

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/action/model/CertificateActionFactory.java
ChrilleSigerhed: If the certificate is a draft, only the sign-resourcelink will be provided.

If the certificate is signed, the SEND_AFTER_SIGN-resourcelink will be provided. This will need to be refactored when we introduce the ability to complement. But lets solve that issue when we introduce that functionality, and go with the simplest implemenation for now?

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/action/model/CertificateActionFactory.java
ChrilleSigerhed: Even tho I dont think its a scenario that can occur, I think you are correct and we should do this check! Good point!

mww-sopra-steria (approved):
Looks good!

PR #184, opened 2024-05-17
INTYGFV-16578: Restricted access for dentist to handle drafts

mhornfeldt:


ChrilleSigerhed:
From my understanding this change will result in that the ```/api/certificatetypeinfo``` response still will contain all certificates from certificate-service, but without the CREATE resourcelink.

Is that a correct assumtion? 

Reason for asking is that I believe the certificates still would be visible for dentists in that case?

mhornfeldt:
> From my understanding this change will result in that the `/api/certificatetypeinfo` response still will contain all certificates from certificate-service, but without the CREATE resourcelink.
> 
> Is that a correct assumtion?
> 
> Reason for asking is that I believe the certificates still would be visible for dentists in that case?

I must've read the code wrong yesterday. I thought CS only returned certificates which could be created in the list, but I see now it returns all of them. I'll reevaluate how to remove it from the list.

PR #187, opened 2024-05-21
INTYGFV-16616: Added endpoint to retrieve messages for certificate.

ChrilleSigerhed:

PR #189, opened 2024-05-23
INTYGFV-16593: Added endpoint for complement certificate

mhornfeldt:


mww-sopra-steria (changes requested):
Looks good! There is some overlap between this PR and #187.

It would be great to sync from main when #187 has been merged and resolve the conflicts on the COMPLEMENT action.

Other than that there is only a minor comment I have. Nice job!

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/action/model/ActionRuleCertificateHasComplement.java
```diff
@@ -0,0 +1,17 @@
+package se.inera.intyg.certificateservice.domain.action.model;
+
+import java.util.Optional;
+import se.inera.intyg.certificateservice.domain.certificate.model.Certificate;
+
+public class ActionRuleCertificateHasComplement implements ActionRule {
```
mww-sopra-steria, line 6: Suggest that you use message.type to decide if the certificate has a complement message or not. You can actual use an already existing method to do that: certificate.messages(MessageType.COMPLEMENT)

I think also it would be good if you checked that at least one of the complement messages isn't handled already or have an answer. If all are answered and/or handled, there is nothing to complement.

mww-sopra-steria (changes requested):
Sorry to keep you waiting! Everything looks great, but I think we have to make some small adjustments to the action rule that checks if there are complement messages. See comment for more information.

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/action/model/ActionRuleCertificateHasComplement.java
```diff
@@ -11,7 +13,10 @@ public boolean evaluate(Optional<Certificate> certificate,
     return certificate.filter(
         value -> value.messages() != null && !value.messages().isEmpty()
             && value.messages().stream()
-            .anyMatch(message -> message.complements() != null && !message.complements().isEmpty())
+            .anyMatch(message ->
+                message.type() == MessageType.COMPLEMENT
```
ChrilleSigerhed, line 17: Could remove the type check and instead use method value.messages(MessageType.COMPLEMENT) to get complements.

ChrilleSigerhed (approved):
Awsome work, looks great!

PR #190, opened 2024-05-24
INTYGFV-16594: Set messages to handle when a certificate with parent relation COMPLEMENT, RENEW or REPLACED is signed

mww-sopra-steria:
Will add integration tests when https://inera.atlassian.net/browse/INTYGFV-16614 has been merged


ChrilleSigerhed (approved):
Awsome!! Well done!

PR #191, opened 2024-05-24
INTYGFV-16594: Added displayName to WorkCapacityType and map it to displayName in cvType when generating xml

mww-sopra-steria:


ChrilleSigerhed (approved):
Nice!

PR #192, opened 2024-05-27
INTYGFV-16622: Implementation of endpoint /{messageId}/exists & /{messageId}/certificate

ChrilleSigerhed:


mww-sopra-steria (changes requested):
Looks good! And good choice to use the GetCertificateDomainService!

What do you think about adding a few integration tests?

mww-sopra-steria (approved):
Nice!

PR #195, opened 2024-05-28
INTYGFV-16590: Added internal api to fetch CertificateMetadataDTO.

ChrilleSigerhed:


mww-sopra-steria (approved):
Looks great! Nice that you decided to refactor the metadata conversion.

Consider naming service and response with "internal" in them, to follow the naming convention for internal api (e.g. GetCertifiateInternalMetadataService). One thing to consider that these services don't really check any access rights and should not be used other than for "internal" api. We could later maybe find a package structure for these.

PR #201, opened 2024-06-03
INTYGFV-16608: Should not be able to send certificate if replaced by signed certificate.

ChrilleSigerhed:


certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/action/model/ActionRuleChildRelationNotSigned.java
```diff
@@ -0,0 +1,24 @@
...
+import se.inera.intyg.certificateservice.domain.certificate.model.Status;
+
+public class ActionRuleChildRelationNotSigned implements ActionRule {
+
+  @Override
+  public boolean evaluate(Optional<Certificate> certificate,
+      Optional<ActionEvaluation> actionEvaluation) {
+    return certificate
+        .map(Certificate::children)
+        .filter(this::notSigned)
```
mww-sopra-steria, line 16: Is it the same rule no matter what type of relation? Thinking about COMPLEMENT (should never happen) and RENEW?

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/action/model/ActionRuleChildRelationNotSigned.java
```diff
@@ -0,0 +1,24 @@
+package se.inera.intyg.certificateservice.domain.action.model;
+
+import java.util.List;
+import java.util.Optional;
+import se.inera.intyg.certificateservice.domain.certificate.model.Certificate;
+import se.inera.intyg.certificateservice.domain.certificate.model.Relation;
+import se.inera.intyg.certificateservice.domain.certificate.model.Status;
+
+public class ActionRuleChildRelationNotSigned implements ActionRule {
```
mww-sopra-steria, line 9: We have the rules today called ActionRuleChildRelationNoMatch and ActionRuleChildRelationMatch. 

Maybe it is time to expand them what statuses to consider. Right now they consider certain relations and have hardcoded what statuses are relevant.

mww-sopra-steria (changes requested):
Nice with the integration tests as well! It definitely works well but lets consider the comments before merge.

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/action/model/ActionRuleChildRelationNotSigned.java
ChrilleSigerhed: Did not consider the effect this will have for other child relations. I verified that this rule does not apply for RENEW, I will make the required updates while considering the comment about utilizing ActionRuleChildRelationNoMatch.

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/action/model/ActionRuleChildRelationNotSigned.java
ChrilleSigerhed: Added allowedStatuses to ActionRuleChildRelationNoMatch to handle statuses more dynamically.

mww-sopra-steria (approved):
Looks good!

PR #202, opened 2024-06-04
INTYGFV-16652: Make lists of embeddables mutable for hibernate to work correctly

mww-sopra-steria:


mikeyOne (approved):

ChrilleSigerhed (approved):
Intressting behaviour! 

Great that you found the cause, well done!

PR #203, opened 2024-06-04
INTYGFV-16631: Only update metadata for certificate if it is a draft and user has right to update the draft

mww-sopra-steria:


ChrilleSigerhed (approved):
Very nice, and cool to see how easy it is for us to evaluate such complex business logic with a simple if-statement (even tho the underlying logic is far more complex ofc) :)

PR #208, opened 2024-06-11
INTYGFV-16664: If certificate doesn't exist, then return an empty list of messages when fetching messages from internal-api.

mww-sopra-steria:


ChrilleSigerhed (approved):
Nice!

PR #209, opened 2024-06-11
INTYGFV-16639:  Added schematron file for FK3226. Refactored tests and moved them into separate schematron package in app layer.

ChrilleSigerhed:


mww-sopra-steria (approved):
I'm impressed! Really nicely done and great job refactoring tests to the app-module.

PR #211, opened 2024-06-13
INTYGFV-16666: Updated IncomingMessageService to handle Questions and Answers of type MessageType.CONTACT & MessageType.OTHER.

ChrilleSigerhed:


certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/message/service/ReceiveAnswerMessageDomainService.java
```diff
@@ -0,0 +1,43 @@
...
+      throw new CertificateActionForbidden(
+          "Not allowed to receive answer on certificate for %s"
+              .formatted(certificate.id().id()),
+          certificate.reasonNotAllowed(CertificateActionType.RECEIVE_ANSWER, Optional.empty())
+      );
+    }
+
+    if (!certificate.isCertificateIssuedOnPatient(message.personId())) {
+      throw new CertificateActionForbidden(
+          "Not allowed to receive complement on certificate for %s, because patient is not matching"
```
mww-sopra-steria, line 33: Change complement to answer

mww-sopra-steria (approved):
Looks great! 

Consider this before merge: By changing the messageType keys we need to wipe the databases in devtest, sit2 and acctest. That might be a problem for sit2 now. Suggest that you sync with Kristina before you decide.

If it is OK to wipe, then go ahead. If it is not OK, then I suggest that you add a new "changeset" and make additions (and changes) to the database there i.e. adding new message type keys and changing the key for ANSWER from 3 -> 5.

PR #212, opened 2024-06-14
INTYGFV-16677: Added restriction by role for certificate action sign

mikeyOne:


mww-sopra-steria (changes requested):
I think it would be good if we used the ActionRuleRole instead of introducing specific rule that can only be used for signing. This will not scale if we want to control different roles for other actions as well.

An ActionSpecification could take a list of roles and when used, it could create an ActionRuleRole with those specified roles.

mikeyOne:
Ok, will try to do that...

mww-sopra-steria (approved):
Looks great!

PR #213, opened 2024-06-17
INTYGFV-16678: Text fixes for fk3226. Tweaks to how the certificate is represented when in signed state (no rules or messages).

mww-sopra-steria:


ChrilleSigerhed (approved):
Wow, great finding that the validations caused the frontend "error"

Good job!

PR #214, opened 2024-06-17
INTYGFV-16682: Added endpoint for getting messages for unit

mhornfeldt:


certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/unit/service/GetUnitMessagesDomainService.java
```diff
@@ -0,0 +1,51 @@
...
+        ).toList();
+
+    return MessagesResponse.builder()
+        .messages(filteredMessagesBasedOnActionEvaluation)
+        .certificates(certificates)
+        .build();
+  }
+
+  private static boolean isReminderOrAnswer(Message message) {
+    return message.type() == MessageType.ANSWER || message.type() == MessageType.REMINDER;
```
ChrilleSigerhed, line 48: In this PR: https://github.com/sklintyg/core/pull/215/files#diff-3bf51e453e5aa08cb8e1d5c1cffe8ba0eae3e72dca3514e20f6d7006e9515cde we introduce a third type that should be ignored, called MISSING. Not sure how this PR should handle that, but im mentioning it so that we are aware of this and handle it. Maybe depending on who merges first? :)

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/unit/service/GetUnitMessagesDomainService.java
```diff
@@ -0,0 +1,51 @@
...
+        .map(Message::certificateId)
+        .distinct()
+        .map(certificateRepository::getById)
+        .filter(certificate -> certificate.allowTo(
+                CertificateActionType.READ,
+                Optional.of(actionEvaluation)
+            )
+        ).toList();
+
+    final var filteredMessagesBasedOnActionEvaluation = messages.stream()
```
ChrilleSigerhed, line 36: I dont really understand why this is needed, since the search we make on line 26 should assert that these are the same?

Maybe an update of the variable name could make the purpose clearer?

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificate/persistence/repository/MessageEntitySpecification.java
```diff
@@ -0,0 +1,34 @@
...
+import se.inera.intyg.certificateservice.infrastructure.certificate.persistence.entity.MessageEntity;
+
+public class MessageEntitySpecification {
+
+  private MessageEntitySpecification() {
+  }
+
+  public static Specification<MessageEntity> sentEqualsAndGreaterThan(LocalDateTime from) {
+    return (root, query, criteriaBuilder) ->
+        criteriaBuilder.greaterThanOrEqualTo(root.get("sent"), from.toLocalDate().atStartOfDay());
```
mhornfeldt, line 16: Maybe this should be moved to WC so filters already are sent start of day? It's necessary to filter correctly when from and to is set to same day as sent date

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/unit/service/GetUnitMessagesService.java
```diff
@@ -0,0 +1,89 @@
...
+            )
+            .toList()
+        )
+        .build();
+  }
+
+  private Certificate getCertificateForMessage(Message message, List<Certificate> certificateList) {
+    return certificateList.stream()
+        .filter(certificate -> certificate.id().id().equals(message.certificateId().id()))
+        .findAny()
```
ChrilleSigerhed, line 85: I was interessted in knowing the difference between findAny and findFirst. From google:

```Java
The findAny() method returns any element from a Stream, while
the findFirst() method returns the first element in a Stream

This is to allow for maximal performance in parallel operations.
```

In this case, this should not matter right? So I think its fine use findAny!

ChrilleSigerhed (changes requested):
Awsome work, I think this looks great!

Left a comment regarding handling of the type MISSING that we will need to handle from another PR. I dont think that should block the merge of this one, but lets discuss on how to handle it before we move forward!

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificate/persistence/repository/MessageEntitySpecification.java
ChrilleSigerhed: I think its fine to handle it in CS, since we are not manipulating the value in other ways than reading the startOfDay wich could be consider as "bussiness logic" that CS should handle.

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/unit/service/GetUnitMessagesDomainService.java
mhornfeldt: 1. Get all messages
2. Get all certificates
3. Filter certificates which user is not allowed to read
4. Filter messages that user is not allowed to read (the messages that belong to a certificate which has been filtered). I think the name "filteredMessagesBasedOnActionEvaluation" describes this.

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/unit/service/GetUnitMessagesDomainService.java
mhornfeldt: It's better if it's in the PR where the enum is introduced. If it's merged first I'll update this.

mhornfeldt:
@ChrilleSigerhed I answered the comments but didn't see any changes requested

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/unit/service/GetUnitMessagesDomainService.java
ChrilleSigerhed: Ah, I see, very good. Thats something that could be easily missed, great that u catched it! 

To me it would be easier to understand if we updated the variable to ```filteredMessagesBasedOnAllowedToRead```, but thats subjective and im fine with you keeping it as it. 

Thanks for clarifying!

ChrilleSigerhed (approved):
Great job, looks great! 

I will handle the adding of MISSING in my PR.

PR #215, opened 2024-06-19
INTYGFV-16669: Implemented app- & domain logic to handle create, send, delete and save for Messages.

ChrilleSigerhed:


mww-sopra-steria (changes requested):
Great job! It looks really good! A few comments, but nothing major. Open to discuss if needed.

mww-sopra-steria (approved):
Looks great!

PR #216, opened 2024-06-25
INTYGFV-16670: Implemented endpoints for create/save, delete and send answer.

ChrilleSigerhed:


certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/message/MessageController.java
```diff
@@ -106,4 +118,25 @@ SendMessageResponse sendMessage(
       @PathVariable("messageId") String messageId) {
     return sendMessageService.send(request, messageId);
   }
+
+  @PostMapping("/{messageId}/saveanswer")
```
mhornfeldt, line 122: Should we have made a AnswerController instead so the endpoint could be /answer/save instead?

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/message/MessageController.java
```diff
@@ -106,4 +118,25 @@ SendMessageResponse sendMessage(
...
+  SaveAnswerResponse saveAnswer(
+      @RequestBody SaveAnswerRequest request,
+      @PathVariable("messageId") String messageId) {
+    return saveAnswerService.save(request, messageId);
+  }
+
+  @PostMapping("/{messageId}/deleteanswer")
```
mhornfeldt, line 129: Should this not be a DeleteMapping?

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/message/service/SaveAnswerService.java
```diff
@@ -0,0 +1,59 @@
...
+        request.getUnit(),
+        request.getCareUnit(),
+        request.getCareProvider()
+    );
+
+    final var message = messageRepository.getById(
+        new MessageId(messageId)
+    );
+
+    final var certificate = certificateRepository.getById(
```
mhornfeldt, line 43: Same question as above

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/message/service/SendAnswerService.java
```diff
@@ -0,0 +1,60 @@
...
+        request.getUnit(),
+        request.getCareUnit(),
+        request.getCareProvider()
+    );
+
+    final var message = messageRepository.getById(
+        new MessageId(messageId)
+    );
+
+    final var certificate = certificateRepository.getById(
```
mhornfeldt, line 44: Same question as above

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/event/model/MessageEventType.java
```diff
@@ -6,6 +6,8 @@ public enum MessageEventType {
   ANSWER_COMPLEMENT("answer-complement-message", Constants.ACTION_TYPE_CREATION,
       "message-sent"),
   SEND_QUESTION("send-question-message", Constants.ACTION_TYPE_CREATION,
+      "message-sent"),
+  SEND_ANSWER("send-anwer-message", Constants.ACTION_TYPE_CREATION,
```
mhornfeldt, line 10: Anwer -> Answer

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/message/model/Answer.java
```diff
@@ -23,4 +23,21 @@ public class Answer {
...
+    this.authoredStaff = staff;
+    this.sent = LocalDateTime.now();
+  }
+
+  public void delete() {
+    this.status = MessageStatus.DELETED_DRAFT;
+  }
+
+  public void save(Staff staff, Content content) {
+    this.content = content;
```
mhornfeldt, line 39: Shouldn't modified be updated?

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/message/service/DeleteAnswerService.java
```diff
@@ -0,0 +1,57 @@
...
+        request.getUnit(),
+        request.getCareUnit(),
+        request.getCareProvider()
+    );
+
+    final var message = messageRepository.getById(
+        new MessageId(messageId)
+    );
+
+    final var certificate = certificateRepository.getById(
```
mhornfeldt, line 42: Should we not get the certificate using the domain service so it's validated that the user is allowed to access the certificate? Or is it enough that the delete action itself is validated in its domain service?

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/message/MessageController.java
ChrilleSigerhed: Hmmm... I kinda like the idea, but im not sure it makes sense. 

Reason for this beeing that the messageId that is provided is not connected to the answer, but to the message. We then later use that id to retrieve the Message and update the answer.

To me thats a good argument to keep it in the MessageController, since the services handle Messages. Do you agree?

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/message/MessageController.java
ChrilleSigerhed: Good point, I also update deleteMessage endpoint and made the required changes in Webcert!

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/message/model/Answer.java
ChrilleSigerhed: Its not done in Webcert and the value is not displayed in frontend, but I think it sounds reasonable to update it. Done!

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/message/service/DeleteAnswerService.java
ChrilleSigerhed: I think its best to let the DeleteAnswerService get the certificate from the infra layer instead of the domain layer. 

Reason for this beeing that the domain service in addition to validate if the user has access, also logs a read event, and that is not correct in this context.

I think its fine to let the DeleteAnswerDomainService determine if the delete action itself is validated or not, and not combine different action validators.

mhornfeldt (approved):

PR #220, opened 2024-06-27
INTYGFV-16671: Extended implementation of handle question to include …

mikeyOne:
…administrative messages


certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/message/service/HandleMessageDomainService.java
```diff
@@ -17,6 +18,45 @@ public class HandleMessageDomainService {
...
+      message.handle();
+    } else {
+      message.unhandle();
+    }
+
+    return messageRepository.save(message);
+  }
+
+  private static void deleteDraftAnswerIfExists(Message message) {
+    if (message.answer() != null && message.answer().status() == MessageStatus.DRAFT) {
```
ChrilleSigerhed, line 53: Just a thought, but for certificates, we are able to just call isDraft() function that returns true if draft, and false if not.

Should implement the same logic in answer?

```java
  public boolean isDraft() {
    return MessageStatus.DRAFT.equals(this.status);
  }
```

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/message/service/SetMessagesToHandleDomainService.java
```diff
@@ -12,6 +13,10 @@ public class SetMessagesToHandleDomainService {
 
   public void handle(List<Message> messages) {
     messages.forEach(Message::handle);
+    messages.stream()
+        .filter(message -> message.answer() != null
+            && message.answer().status() == MessageStatus.DRAFT)
```
ChrilleSigerhed, line 18: Use isDraft function if we seem fit to implement it!

ChrilleSigerhed (approved):
Very good job, I think this looks great!

Left a minor suggestion regarding moving the status check into the Answer class, just so that we follow the same pattern as for Certificate!

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/message/service/HandleMessageDomainService.java
mikeyOne: done

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/message/service/SetMessagesToHandleDomainService.java
mikeyOne: done

PR #226, opened 2024-07-02
INTYGFV-16694: Should not include CertificateDataElement if certificate is not draft and element data is null.

ChrilleSigerhed:


mikaelaromild1 (approved):
Looks good :)

mikeyOne (approved):
looks good

PR #229, opened 2024-07-02
INTYGFV-16707: Added funktionsnedsattning to FK7809

mhornfeldt:


certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/fk7809/CodeSystemFunktionsnedsattning.java
```diff
@@ -0,0 +1,65 @@
+package se.inera.intyg.certificateservice.infrastructure.certificatemodel.fk7809;
+
+import se.inera.intyg.certificateservice.domain.common.model.Code;
+
+public class CodeSystemFunktionsnedsattning {
+
+  public static final String CODE_SYSTEM = "FUNKTIONSNEDSATTNING";
```
ChrilleSigerhed, line 7: Since these codes does not have a codesystem realated to them, does it make sense to set it to null instead?

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/fk7809/CertificateModelFactoryFK7809.java
```diff
@@ -162,9 +177,61 @@ public class CertificateModelFactoryFK7809 implements CertificateModelFactory {
...
+  private static final FieldId FUNKTIONSNEDSATTNING_INTELLEKTUELL_FUNKTION_ID = new FieldId("8.2");
+  private static final FieldId FUNKTIONSNEDSATTNING_KOMMUNIKATION_SOCIAL_INTERAKTION_ID = new FieldId(
+      "9.2");
+  private static final FieldId FUNKTIONSNEDSATTNING_UPPMAKRMSAHET_ID = new FieldId("10.2");
+  private static final FieldId FUNKTIONSNEDSATTNING_PSYKISK_FUNKTION_ID = new FieldId("11.2");
+  private static final FieldId FUNKTIONSNEDSATTNING_HORSELFUNKTION_ID = new FieldId("48.2");
+  private static final FieldId FUNKTIONSNEDSATTNING_SYNFUNKTION_ID = new FieldId("49.2");
+  private static final FieldId FUNKTIONSNEDSATTNING_SINNESFUNKTION_ID = new FieldId("12.2");
+  private static final FieldId FUNKTIONSNEDSATTNING_KOORDINATION_ID = new FieldId("13.2");
+  private static final FieldId FUNKTIONSNEDSATTNING_ANNAN_KROPPSILIG_FUNKTION_ID = new FieldId(
```
ChrilleSigerhed, line 193: I was first confused why I could not find a 8.2 value in the XML, then I realised its a fieldId and not elementId.

To me it would make more sense to call these "json-id:s" something like "annanKroppsiligFunktion" instead of a traditional questionId. But this is very subjective, and I dont know if we have established a pattern for this yet. If you agree with me, update it... if not, dont mind this comment! :)

ChrilleSigerhed (approved):
Awsome work, looks great. I like that you also added a XmlGenerator for CodeList even tho its not beeing used yet, right? :)

The code reads good and follows the pattern we already established, really well done. I added some comments, but I let you decide if its worth updateing or not, im also open for discussion if thats what u prefer!

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/fk7809/CodeSystemFunktionsnedsattning.java
mhornfeldt: I think having it symbolizes it's a code system we set up and not FK

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/fk7809/CertificateModelFactoryFK7809.java
mhornfeldt: I'll leave them as they are and we can refactor if we decide to set it as a rule for CS

PR #230, opened 2024-07-03
INTYGFV-16641: Refactering of integration tests for easier usage. Add…

mikeyOne:
…ed tests for fk3226 and fk7809.

### DIFF
--- a/certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/fk7809/CertificateModelFactoryFK7809.java
+++ b/certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/fk7809/CertificateModelFactoryFK7809.java
@@ -71,7 +71,7 @@ import se.inera.intyg.certificateservice.infrastructure.certificatemodel.fk3226.
 public class CertificateModelFactoryFK7809 implements CertificateModelFactory {
 
   private static final String GENERAL_LABEL_FUNKTIONSNEDSATTNING = "Beskriv funktionsnedsättningen, om möjligt med grad. Ange även eventuella undersökningsfynd.";
-  @Value("${certificate.model.fk7210.v1_0.active.from}")
+  @Value("${certificate.model.fk7809.v1_0.active.from}")
   private LocalDateTime activeFrom;
 
   private final DiagnosisCodeRepository diagnosisCodeRepositoy;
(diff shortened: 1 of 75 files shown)


certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/fk7809/CertificateModelFactoryFK7809.java
```diff
@@ -71,7 +71,7 @@
 public class CertificateModelFactoryFK7809 implements CertificateModelFactory {
 
   private static final String GENERAL_LABEL_FUNKTIONSNEDSATTNING = "Beskriv funktionsnedsättningen, om möjligt med grad. Ange även eventuella undersökningsfynd.";
-  @Value("${certificate.model.fk7210.v1_0.active.from}")
+  @Value("${certificate.model.fk7809.v1_0.active.from}")
```
ChrilleSigerhed, line 74: Ops! Nice catch! :D

certificate-service/app/src/testFixtures/java/se/inera/intyg/certificateservice/application/testdata/TestDataCommonUserDTO.java
```diff
@@ -103,6 +117,26 @@ public static UserDTOBuilder alvaVardadministratorDtoBuilder() {
         .healthCareProfessionalLicence(Collections.emptyList());
   }
 
+  public static UserDTOBuilder annaSJukskoterskaDtoBuilder() {
```
ChrilleSigerhed, line 120: Maybe fix the capital "J"

ChrilleSigerhed (changes requested):
Wow, excellent work with this. Now we have established a clear and good pattern for how we will implement integrationtests in the future, I really like what you did here!

I have one small request, and that is that we maybe could update the project structure and place the integrationtests in different packages so that they are more easily located. For example maybe tests regarding messages, like MessagesIT, should live in an folder called messages.

What do you think about that?

ChrilleSigerhed (approved):
Awsome work! Well done! Will handle filltype.MAXIMAL later as discussed.

certificate-service/app/src/testFixtures/java/se/inera/intyg/certificateservice/application/testdata/TestDataCommonUserDTO.java
mikeyOne: done

PR #231, opened 2024-07-04
INTYGFV-16708: Added schematron for fk7809

mhornfeldt:


certificate-service/app/src/main/resources/fk7809/schematron/lumek.v1.sch
```diff
@@ -0,0 +1,413 @@
...
+  <iso:pattern id="q1.3">
+    <iso:rule context="//gn:delsvar[@id='1.3']">
+      <iso:extends rule="non-empty-string"/>
+      <iso:assert test="count(../gn:delsvar[@id='1.1']/tp:cv/tp:code[normalize-space(.) != 'ANNAT']) = 0">
+        Om 'Typ av grund för MU' inte är 'Annat' så får 'Ange vad annat är' inte anges.
+      </iso:assert>
+    </iso:rule>
+  </iso:pattern>
+
+  <iso:pattern id="q1.1-1.4">
```
mhornfeldt, line 117: Should this be 1.1-1.4 or 1.1-1.3?

ChrilleSigerhed: 1.1-1.3 makes sense since the pattern is testing delsvar 1.3. Seems like we already defined a 1.1-1.4 on line 134 to test the 1.4 delsvar!

ChrilleSigerhed (approved):
Impressive work, and the test suite to back it up is great! 

Also great usage of regex to handle different types of 'Funktionsnedsättning'

Consider updating the id we already discussed before merging!

certificate-service/app/src/main/resources/fk7809/schematron/lumek.v1.sch
mhornfeldt: Okay but isn't it weird then if the test for 1.4 includes 1.3?

mikeyOne:
Agree, very impressive work!

PR #233, opened 2024-07-04
INTYGFV-16641: Implemented maximal filltype for certificates

mikeyOne:


ChrilleSigerhed (approved):
Awsome work, looks great! I really like the way you introduced the TestabilityFIllCertificateUtil and usage of it.

PR #238, opened 2024-07-09
INTYGFV-16699: Updated text and removed unnecessary white space in he…

mikeyOne:
…lp text


certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/common/CodeSystemKvFkmu0001.java
```diff
@@ -15,7 +15,7 @@ public class CodeSystemKvFkmu0001 {
   public static final Code JOURNALUPGIFTER = new Code(
       "JOURNALUPPGIFTER",
       CODE_SYSTEM,
-      "journaluppgifter från och med"
```
mhornfeldt, line 18: We need to verify this text change since it will be applied on fk7809 also. Rebecka asked us to update it for fk7809 last week.

mhornfeldt (approved):

PR #241, opened 2024-07-09
INTYGFV-16722: Implemented endpoint for forward certificate.

ChrilleSigerhed:


ChrilleSigerhed:
Also added MessageType to FK7809 so that "Avstämningsmöte" is not an viable option.

PR #242, opened 2024-07-10
INTYGFV-16704: Test indicated certificates should not be shown for citizen on 1177

mhornfeldt:


certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/citizen/service/GetCitizenCertificateListService.java
```diff
@@ -39,6 +39,10 @@ public GetCitizenCertificateListResponse get(
     return GetCitizenCertificateListResponse.builder()
         .citizenCertificates(certificates.stream()
             .filter(certificate -> certificate.certificateModel().availableForCitizen())
+            .filter(
```
mhornfeldt, line 42: Would this belong better as a function of Certificate or Patient object?

ChrilleSigerhed: I think it makes sense to move the logic to the Patient object!

mhornfeldt: Doesn't seem like we've done it for protected person or deceased so maybe it's overkill since it's only a boolean

mhornfeldt: Also doesn't seem like it can be null

ChrilleSigerhed (approved):
Nice!

PR #244, opened 2024-07-10
INTYGFV-16722: Refactor so that available messages are set from the model using MessageActionSpecification. Updated logic for handle message so that correct behaviour is displayed. Added certificate and message package to action package.

ChrilleSigerhed:
Messages from recipient with answers should be displayed as handled, and not able to uncheck it.
image


certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/action/certificate/model/CertificateActionSign.java
```diff
@@ -67,4 +68,40 @@ public String getDescription(Optional<Certificate> optionalCertificate) {
         )
         .orElse(SIGN_DESCRIPTION);
   }
+
+  public static class ActionRuleBlockTestIndicatedPerson implements ActionRule {
```
mikaelaromild1, line 72: Move this to a separate class :)

mikaelaromild1 (changes requested):

mikaelaromild1 (approved):
Well done :)

PR #245, opened 2024-07-11
INTYGFV-16702: Added rule UserHasAccessScope to determine if user has correct access scope for functions.

ChrilleSigerhed:


mikaelaromild1 (approved):
Nice! This is subjective but I personally preferred the name ActionRuleIntegrated over ActionRuleUserHasAccessScope. But on the other hand this term is more general so maybe it´s a better alternative :)

PR #247, opened 2024-07-11
INTYGFV-16625: Added ResponsibleHospName  to CertificateMetadata

mikaelaromild1:


certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/certificate/service/converter/CertificateMetadataConverter.java
```diff
@@ -114,6 +114,9 @@ public CertificateMetadataDTO convert(Certificate certificate) {
                 .toList()
                 : null
         )
+        .responsibleHospName(certificate.certificateMetaData().responsibleHospName() != null
```
mhornfeldt, line 117: Do we want to use "responsibleHospName" or name it something different? I remember we removed the term hosp when working with intyg-proxy.

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/common/ActionEvaluationFactory.java
```diff
@@ -76,6 +77,7 @@ public ActionEvaluation create(PatientDTO patient, UserDTO user, UnitDTO unit, U
                         .map(HealthCareProfessionalLicence::new)
                         .toList()
                 )
+                .responsibleHospName(new ResponsibleHospName(user.getResponsibleHospName()))
```
mhornfeldt, line 80: Do we want to use "responsibleHospName" or name it something different? I remember we removed the term hosp when working with intyg-proxy.

mhornfeldt (approved):

ChrilleSigerhed (approved):
Looks great! Well done!

image

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/common/ActionEvaluationFactory.java
ChrilleSigerhed: Maybe we should update it to ResponsibleIssuer?

PR #249, opened 2024-07-11
INTYGFV-16729: Added isEmpty for ElementValue

mhornfeldt:


ChrilleSigerhed (changes requested):
Very good, I like the idea very much!

I think we should add some tests to verify the isEmpty-function in each type before merging

ChrilleSigerhed (approved):
Nice!

PR #252, opened 2024-07-15
INTYGFV-16650: Implemented CertificateAction responsible issuer

mikaelaromild1:


ChrilleSigerhed (changes requested):
Looks great, well done!

Should we add some integrationtests to verify this behaviour?

ChrilleSigerhed (approved):
Awsome, good job!

PR #255, opened 2024-07-18
INTYGFV-16737: Refactored pdf generator and moved certificate type specific code to certificate model package

mikaelaromild1:


certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificatemodel/model/PdfQuestionField.java
```diff
@@ -0,0 +1,13 @@
...
+
+import lombok.Builder;
+import lombok.Value;
+
+@Value
+@Builder
+public class PdfQuestionField {
+
+  ElementId questionId;
+  String pdfFieldId;
```
ChrilleSigerhed, line 11: Should we use the type FieldId here instead? We try to work with objects if possible in the domain layer!

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificatemodel/model/PdfSpecification.java
```diff
@@ -0,0 +1,17 @@
...
+
+@Value
+@Builder
+public class PdfSpecification {
+
+  CertificateType certificateType;
+  int mcid;
+  int signatureWithAddressTagIndex;
+  int signatureWithoutAddressTagIndex;
+  String patiendIdFieldId;
```
ChrilleSigerhed, line 15: Should we use the type FieldId here instead?

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificatemodel/model/PdfSpecification.java
```diff
@@ -0,0 +1,17 @@
...
+import java.util.List;
+import lombok.Builder;
+import lombok.Value;
+
+@Value
+@Builder
+public class PdfSpecification {
+
+  CertificateType certificateType;
+  int mcid;
```
ChrilleSigerhed, line 12: Maybe this can be its own type? Maybe a record called Mcid(int value)?

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificatemodel/model/PdfSpecification.java
```diff
@@ -0,0 +1,17 @@
...
+  int signatureWithAddressTagIndex;
```
ChrilleSigerhed, line 13: signatureWithAddressTagIndex & signatureWithoutAddressTagIndex could both the records of type TagIndex?

certificate-service/pdfbox-generator/src/main/java/se/inera/intyg/certificateservice/pdfboxgenerator/pdf/value/PdfElementValueGenerator.java
```diff
@@ -1,14 +1,40 @@
...
-public interface PdfElementValueGenerator {
+@Service
+@RequiredArgsConstructor
+public class PdfElementValueGenerator {
 
-  List<PdfField> generate(Certificate certificate, ElementId id, String fieldId)
-      throws IOException;
+  private final List<PdfElementValue> pdfElementValues;
 
+  public List<PdfField> getFields(Certificate certificate) {
```
ChrilleSigerhed, line 17: Should this method be called generate, and the already defined generate method maybe should be called getFields?

I think it better describes both methods?

certificate-service/pdfbox-generator/src/main/java/se/inera/intyg/certificateservice/pdfboxgenerator/pdf/value/PdfElementValueGenerator.java
```diff
@@ -1,14 +1,40 @@
...
+    return certificate.certificateModel().pdfSpecification().pdfQuestionFields()
+        .stream()
+        .map(
+            questions -> generate(certificate, questions.questionId(), questions.pdfFieldId(),
```
ChrilleSigerhed, line 21: Since questions in this contex is a PdfQuestionField, should we rename the variable to pdfQuestion instead? Its a minor thing but I think it communicates better what we are working with. Calling it questions might indicate its a list.

certificate-service/pdfbox-generator/src/test/java/se/inera/intyg/certificateservice/pdfboxgenerator/pdf/value/PdfElementValueGeneratorTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import se.inera.intyg.certificateservice.domain.certificatemodel.model.PdfValueType;
+
+@ExtendWith(MockitoExtension.class)
+class PdfElementValueGeneratorTest {
```
ChrilleSigerhed, line 24: I think the tests you have added are great, and clearly verifies that correct valueConverter is called. But im missing a test that verifies that we are actually getting a List<PdfField> returned.

Could you please add that aswell? If we changed the logic now so that this service returns a list of another object, the tests would not catch that!

ChrilleSigerhed (changes requested):
Great work, I think this improved the way that we handle PDF's greatly!

Added some comments to consider before merging, mainly types and name of variables that needs to be updated.

Well done!

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificatemodel/model/PdfQuestionField.java
mikaelaromild1: Changing this to PdfFieldId as discussed since they are two different things.

ChrilleSigerhed (approved):
Impressive work, this looks awsome!

PR #256, opened 2024-07-26
INTYGFV-16738: Implemented PDF for FK3226, added new generators for PdfValueTypes…

mikaelaromild1:
… and introduced QuestionConfiguration to PdfSpecification


certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/fk3226/CertificateModelFactoryFK3226.java
```diff
@@ -150,8 +161,137 @@ public class CertificateModelFactoryFK3226 implements CertificateModelFactory {
   public static final SchematronPath SCHEMATRON_PATH = new SchematronPath(
       "fk3226/schematron/lunsp.v1.sch");
 
-  public static final String PDF_FK_3226_PDF = "fk3226/pdf/fk3226_v1.pdf";
-  public static final String PDF_NO_ADDRESS_FK_7472_PDF = "fk3226/pdf/fk3226_v1_no_address.pdf";
+  public static final String PDF_FK_3226_PDF = "fk3226/pdf/fk3226_v1_no_address.pdf"; //TODO: update when new template has been provided
+  public static final String PDF_NO_ADDRESS_FK_3226_PDF = "fk3226/pdf/fk3226_v1_no_address.pdf";
+  public static final Mcid PDF_MCID = new Mcid(100);
```
mikaelaromild1, line 166: From what I can tell this mcid works, but I'm not 100% sure.

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/fk3226/CertificateModelFactoryFK3226.java
```diff
@@ -150,8 +161,137 @@ public class CertificateModelFactoryFK3226 implements CertificateModelFactory {
...
+  private static final int PDF_SIGNATURE_PAGE_INDEX = 1;
```
mikaelaromild1, line 167: Maybe it's better to move all PDF-constants to a separate class since there are so many. But since we discussed that all certificate type specific code should be placed in or as close to the model as possible i decided to let them stay here for now.

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/fk3226/CertificateModelFactoryFK3226.java
mww-sopra-steria: We will test it and modify it later if necessary.

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/fk3226/CertificateModelFactoryFK3226.java
mww-sopra-steria: I think you are on to something. Our factory-class has grown quite big and it could do with some refactoring, but it isn't necessary to do it in this PR.

certificate-service/pdfbox-generator/src/main/java/se/inera/intyg/certificateservice/pdfboxgenerator/pdf/text/PdfTextGenerator.java
```diff
@@ -42,9 +42,10 @@ private void addText(
       Float offsetY,
       boolean isBold,
       int mcid,
-      PDStructureElement section
+      PDStructureElement section,
+      int pageIndex
   ) throws IOException {
-    final var page = pdf.getPage(0);
+    final var page = pdf.getPage(pageIndex);
```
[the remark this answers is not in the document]
mww-sopra-steria: I get that. But I think in this instans it could be OK that it is specified for each certificate type. This can vary between different types and when it comes to other that FK certificates it could look very different.

mww-sopra-steria (approved):
It works like a charm! Really good job!

I will take a look at some of the points your raised in the PR (and in your email) and look to refactor it a bit. The structure you have established is really good and I think we can take it further by moving more logic from the PDF-module to the model. Maybe also use some of the structure that we already have in the module to do the mapping to pdf-fields.

It would be great if we can let the PDF-module create the PDF and the model to provide the values, so we don't have to have mapping logic for different type of values in the PDF-module.

PR #259, opened 2024-08-07
INTYGFV-16752: Added pdf logic for fk7809 and fixed logic related to multiple pages pdf

mhornfeldt:


mww-sopra-steria (approved):
Looks great! Well done!

PR #260, opened 2024-08-08
INTYGFV-16753: Added text overflow logic for pdf

mhornfeldt:


mww-sopra-steria (approved):
Looks good! I like the way you have implemented this and it's very simple to use the functionality from the models perspective. Well done!

PR #267, opened 2024-08-19
INTYGFV-16771: Added logic for getting events for certificate

mhornfeldt:


certificate-service/integration-test/src/test/java/se/inera/intyg/certificateservice/integrationtest/GetCertificateEventsIT.java
```diff
@@ -0,0 +1,212 @@
...
+            .user(ALVA_VARDADMINISTRATOR_DTO)
+            .build(),
+        certificateId(testCertificates)
+    );
+
+    assertEquals(403, response.getStatusCode().value());
+  }
+
+  @Test
+  @DisplayName("Om intyget är utfärdat på en patient som har skyddade personuppgifter skall dess händelser returneras")
```
ChrilleSigerhed, line 192: Should we add the prefix Läkare to the DisplayName?

ChrilleSigerhed (changes requested):
Awsome work, the solution is clean and easy to follow, impressive!

Im glad you were able to find a solution and build the certificate events from the data already present to us from the certificate.

I left some minor comments to consider.

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificate/service/GetCertificateEventsOfTypeDomainService.java
```diff
@@ -0,0 +1,146 @@
...
+        return certificate.sent() != null
+            ? List.of(eventFromTimestamp(certificate, certificate.sent().sentAt(), type))
+            : Collections.emptyList();
+      }
+      case INCOMING_MESSAGE -> {
+        return certificate.messages()
+            .stream()
+            .filter(message -> message.type() != MessageType.ANSWER
+                && message.type() != MessageType.COMPLEMENT)
+            .filter(message -> message.author().author().equals("FK"))
```
ChrilleSigerhed, line 32: Maybe the author class can be responsible for checking if its from FK?

```java
  private static final String FK = "FK";

  public boolean fromFK() {
    return this.name().equals(FK);
  }
```

mhornfeldt: I can do that, but it's not gonna work with name since it's transformed to Försäkringskassan. I'll use author.

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificate/service/GetCertificateEventsOfTypeDomainService.java
```diff
@@ -0,0 +1,146 @@
...
+      }
+      case SENT -> {
+        return certificate.sent() != null
+            ? List.of(eventFromTimestamp(certificate, certificate.sent().sentAt(), type))
+            : Collections.emptyList();
+      }
+      case INCOMING_MESSAGE -> {
+        return certificate.messages()
+            .stream()
+            .filter(message -> message.type() != MessageType.ANSWER
```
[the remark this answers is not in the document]
mhornfeldt: Feels a bit overkill since it's just checking one value. If two types matched to the same kind I think it would add value. I looked and we've only done it on draft before which I'm guessing is because draft will have several statuses when we introduce locked. I'll leave this as it is.

ChrilleSigerhed (approved):
Awsome! Great job!

PR #274, opened 2024-08-25
INTYGFV-16793: Removed copy parameter from actions that do not have t…

mhornfeldt:
…hat requirement


mww-sopra-steria (approved):
Great that you fixed the other places where the rule was used incorrectly!

PR #279, opened 2024-09-02
INTYGFV-16795: Updated ActionRules to match requirement specification

mikaelaromild1:


certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/action/certificate/model/CertificateActionFactory.java
```diff
@@ -127,7 +130,9 @@ public static CertificateAction create(CertificateActionSpecification actionSpec
                   new ActionRuleChildRelationNoMatch(List.of(RelationType.REPLACE),
                       List.of(Status.REVOKED)),
                   new ActionRuleUserNotBlocked(),
-                  new ActionRuleNoComplementMessages()
+                  new ActionRuleNoComplementMessages(),
+                  new ActionRuleProtectedPerson(),
+                  new ActionRuleInactiveUnit()
```
mww-sopra-steria, line 135: new ActionRulePatientAlive() should also be considered when replacing. 

Suggest we don't include it in REPLACE_CONTINUE as the draft has already been created and there is no rule for not allowing the user to sign a draft even if the patient is deceased.

mww-sopra-steria (changes requested):
So good that you validated the rules and also found and fixed a quite few of them! Well done! 

I found one thing that should be added to the REPLACE action after double checking the requirements.

mww-sopra-steria (approved):
Looks good! Well done!

PR #280, opened 2024-09-03
K1J-85: Adjust multiline textfield height

SamuelTilly:


mww-sopra-steria (approved):
Looks good! Well done!

PR #285, opened 2024-09-04
K1J-149: Shall include readyForSignedBy when building StaffRepository. Removed readyForSign as standard attribute from FK7210 builder.

ChrilleSigerhed:


mww-sopra-steria (approved):
Looks good!

PR #290, opened 2024-09-04
K1J-150: Added integration tests for ready for sign

mhornfeldt:


certificate-service/integration-test/src/test/java/se/inera/intyg/certificateservice/integrationtest/CertificateReadyForSignIT.java
```diff
@@ -0,0 +1,90 @@
...
+import se.inera.intyg.certificateservice.application.common.dto.AccessScopeTypeDTO;
+
+public abstract class CertificateReadyForSignIT extends BaseIntegrationIT {
+
+  protected abstract String type();
+
+  protected abstract String typeVersion();
+
+  @Test
+  @DisplayName("Vårdadministratör - Om användaren är en vårdadministratör ska intyget gå att markera som redo för signering")
```
ChrilleSigerhed, line 23: I think it would be great if we clearified that its an deep integrated care admin in the display name. Also, maybe instead of saying "intyget" we should refeer to it as "utkast" to better communicate the state of the certificate?

Maybe we should also add a test where the care admin is not integrated and assert that an error is thrown!

ChrilleSigerhed (approved):
Nice!

I added some suggestions of improvements that could be done before merging!

PR #292, opened 2024-09-10
K1J-171: Added rule to block renew if certificate has been replaced or complemented.

mww-sopra-steria:


ChrilleSigerhed (approved):
Nice!

PR #293, opened 2024-09-11
K1J-175: Should not show created event if created by replacing or complementing

mikaelaromild1:


mww-sopra-steria (approved):
Looks good!

PR #296, opened 2024-09-13
K1J-157: Truncated diagnosis description and fixed position of text on overflow sheet

mikaelaromild1:


certificate-service/pdfbox-generator/src/main/java/se/inera/intyg/certificateservice/pdfboxgenerator/pdf/CertificatePdfFillService.java
```diff
@@ -194,12 +194,18 @@ private void setFieldValue(PDAcroForm acroForm, PdfField field)
       final var extractedField = acroForm.getField(field.getId());
       validateField(field.getId(), extractedField);
 
-      if (extractedField instanceof PDTextField textField && textField.isMultiline()) {
-        final var textAppearance = new TextfieldAppearance(textField);
+      final var append = field.getAppend() != null && field.getAppend();
+
+      if (extractedField instanceof PDTextField textField && field.getAppearance() != null) {
+        textField.setDefaultAppearance(field.getAppearance());
```
mikaelaromild1, line 200: Not sure if or how this could be tested.

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/fk3226/elements/QuestionDiagnos.java
```diff
@@ -154,6 +154,8 @@ public static ElementSpecification questionDiagnos(
         .pdfConfiguration(
             PdfConfigurationDiagnoses.builder()
                 .prefix(PDF_DIAGNOSIS_FIELD_ID_PREFIX)
+                .maxLength(172)
```
mhornfeldt, line 157: Wasn't the asked solution that we didn't have a max limit and instead decided if the text fit in the field dynamically?

certificate-service/pdfbox-generator/src/test/java/se/inera/intyg/certificateservice/pdfboxgenerator/pdf/value/PdfValueGeneratorUtilTest.java
```diff
@@ -22,7 +22,16 @@ void shouldSplitTextUsingShortInformationTextIfLimitIsUnder25() {
...
+    assertEquals(expected, PdfValueGeneratorUtil.splitByLimit(10, value, true));
   }
 
-}
\ No newline at end of file
+  @Test
+  void shouldSplitTextUsingShortInformationTextIfSplitTextShouldNotContinueOnOverFLowSheet() {
+    final var value = "This is the value if was planning on testing for splitting text.";
+    final var expected = List.of("This is ...",
+        "... the value if was planning on testing for splitting text.\n");
```
mhornfeldt, line 32: Do we really want a \n if not added on an overflow sheet?

certificate-service/pdfbox-generator/src/main/java/se/inera/intyg/certificateservice/pdfboxgenerator/pdf/value/PdfValueGeneratorUtil.java
```diff
@@ -11,9 +11,13 @@ private PdfValueGeneratorUtil() {
     throw new IllegalStateException("Utility class");
   }
 
-  public static List<String> splitByLimit(Integer limit, String s) {
+  public static List<String> splitByLimit(Integer limit, String s,
+      boolean addOverflowOnContinuationSheet) {
```
mhornfeldt, line 15: Instead of adding this logic to each of the values, isn't it better than in the diagnosis question for 3226 and 7809 we don't submit an overflow sheet, and the logic if added on an overflow sheet will only apply if the sheet is available

mhornfeldt (changes requested):

mhornfeldt:
Shouldn't we build a solution that works for both diagnosis and text field? If the PdfConfiguration is missing a overflow sheet it should be truncated with "..." and not moved. If an overflow sheet is added in the config it should be truncated with "... Se fortsättningsblad!" and moved. Also, I understood it from forvaltningen that they wanted us to only truncate the diagnosis field when it was full, not after a specific max length.

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/fk3226/elements/QuestionDiagnos.java
mikaelaromild1: Forvaltningen has clarified that they are satisfied with the current solution, i.e. a max limit for text fields.

certificate-service/pdfbox-generator/src/main/java/se/inera/intyg/certificateservice/pdfboxgenerator/pdf/CertificatePdfFillService.java
mhornfeldt: Maybe you can get access to the field in the test and check that the appearance has been set/not set?

certificate-service/pdfbox-generator/src/test/java/se/inera/intyg/certificateservice/pdfboxgenerator/pdf/value/PdfDiagnosisListValueGeneratorTest.java
```diff
@@ -264,5 +264,73 @@ void shouldSplitValueIfOverMaxLength() {
...
+                      .build()
+              )
+          )
+          .build();
+
+      final var result = pdfDiagnosisListValueGenerator.generate(elementSpecification,
+          elementValue);
+
+      assertEquals(expected, result);
+    }
```
mhornfeldt, line 334: Shouldn't we test that appearance is set?

mhornfeldt (commented):
Added some comments

PR #297, opened 2024-09-13
K1J-205: Updated mapping of diagnosis and related schematron. Also updated tests and reformatted the schematron-files.

ChrilleSigerhed:


mww-sopra-steria (approved):
Looks good! Well done!

PR #304, opened 2024-09-20
K1J-167-1: Removed integration tests for RESPONSIBLE_ISSUER

mikaelaromild1:


mhornfeldt (approved):
Is there any other logic that should be tested instead for responsible issuer in an IT test?

mikaelaromild1:
No, not since there is no logic related to this in CS anymore. User origin is used in frontend to display responsible issuer now instead of resource links/certificate actions

PR #305, opened 2024-10-03
K1J-335: Information that messages are not available should be displayed for not sent certificates

mikaelaromild1:


mhornfeldt (approved):
Nice!

PR #311, opened 2024-10-10
K1J-394: Should send certificate after sign if complemented

mhornfeldt:


mww-sopra-steria (changes requested):
As discussed we should handle this with "SEND_AFTER_SIGN"

ChrilleSigerhed (approved):
Nice, good that we valide the logic with a new integrationtest!

PR #312, opened 2024-10-10
K1J-389: Logging in json-format in alignment with ECS. Performance logging on external rest calls. Handle trace.id, span.id and session.id.

mww-sopra-steria:


ChrilleSigerhed (approved):
Awsome, great work!

PR #315, opened 2024-10-11
K1J-383: Value should not be static for logical address for recipient

mhornfeldt:


mww-sopra-steria (changes requested):
This PR adds konfiguration outside of the certificate type model. Could we instead than adding it as a configuration in clinicalprocess-v4 add it to the model?

This will make sure that all configuration related to certificate types will be contained in one place.

mhornfeldt:
> This PR adds konfiguration outside of the certificate type model. Could we instead than adding it as a configuration in clinicalprocess-v4 add it to the model?
> 
> This will make sure that all configuration related to certificate types will be contained in one place.

Add it to the model but the value will still be set from config?

mww-sopra-steria:
> > This PR adds konfiguration outside of the certificate type model. Could we instead than adding it as a configuration in clinicalprocess-v4 add it to the model?
> > This will make sure that all configuration related to certificate types will be contained in one place.
> 
> Add it to the model but the value will still be set from config?

Yes, that would be good. Maybe att it to the Recipient that is part of the domain.

ChrilleSigerhed (approved):
Looks great, well done!

PR #318, opened 2024-10-12
K1J-386: Update to java 21, springboot 3.3.4, gradle 8.10.2 and update third party dependencies

mww-sopra-steria:


mikeyOne (approved):
Approved.

To get rid of a few 'deprecated'-warnings in build.gradle you could have a look at sonar and dependency check properties in for example common and infra where these have been slightly updated.

I have also found that not applying the sonar and dependency-check plugins in all subprojects gives the same result in the aggragated reports that we use, while it appears to reduce the time for analysis in for example our sonar pipelines. (Based on what tasks are available in subrojects of "the old apps" this appears to be how they were used when applied by the gradle plugin).

PR #319, opened 2024-10-12
K1J-390: Performance logging for api calls to certificate-service. Handle trace.id, span.id and session.id.

mww-sopra-steria:


mikeyOne (approved):
Looks great!

PR #320, opened 2024-10-15
K1J-421: Set enabled == true on available functions for citizens.

mww-sopra-steria:


mhornfeldt (approved):
Nice. Is it possible to add as a default value so we don' get this bug again?

mww-sopra-steria:
> Nice. Is it possible to add as a default value so we don' get this bug again?

Good idea and I been thinking a little about it. Conclusion: No matter if we have it as default value or not, we should have unit tests that verifies that the data is constructed correctly.

PR #321, opened 2024-10-15
K1J-403: Midwife and Nurse should be able to forward & mark certificates ready for sign. Midwife and nurse should also be able to send like care admin.

ChrilleSigerhed:


erikagrensopra (approved):
Looks good!

mikaelaromild1:
Shouldn't you only be able to mark certificate redy for sign and forward drafts if you do not have the right to sign? Maybe we need to set allowed roles in each certificate model instead, similar to SIGN or SEND.

According to the requeirements specification in Confluence, a nurse or midwife gets the same rights as a care_admin only if they do not have the right to sign.

mikaelaromild1 (changes requested):
See comment

mikaelaromild1 (approved):
Well done :)

PR #322, opened 2024-10-16
K1J-384: SvarPa should be set if parent has complement message

mhornfeldt:


certificate-service/clinicalprocess-certificate-v4/src/main/java/se/inera/intyg/certificateservice/infrastructure/clinicalprocesscertificatev4/certificate/XmlGeneratorCertificateV4.java
```diff
@@ -83,7 +84,7 @@ private MeddelandeReferens svarPa(Certificate certificate) {
         return null;
       }
 
-      final var complement = unhandledComplements.get(unhandledComplements.size() - 1);
+      final var complement = unhandledComplements.getLast();
```
ChrilleSigerhed, line 87: Ohlala, nice feature!

PR #323, opened 2024-10-16
K1J-421: Added endpoint for sending certificates as citizen.

ChrilleSigerhed:


certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificate/model/Certificate.java
```diff
@@ -236,6 +236,7 @@ private void throwIfConcurrentModification(Revision revision, String operation,
   }
 
   public void send(ActionEvaluation actionEvaluation) {
+    final var optionalActionEvaluation = Optional.ofNullable(actionEvaluation);
```
mww-sopra-steria, line 239: Thinking that it would make sense to add a new send method instead of using the current one and passing "null" for actionEvaluation.

Could be called send(), but could be hard to understand when to use which... so maybe something like sendByCitizen() or similar.

mww-sopra-steria (changes requested):
Looks really good! Well done! I have a small suggestion on the change of Certificate in the domain.

mww-sopra-steria (approved):
Perfecto!

PR #326, opened 2024-10-17
K1J-335: Information that messages are not available should only be displayed for signed certificates

mikaelaromild1:


ChrilleSigerhed (approved):
Nice!

PR #329, opened 2024-10-28
K1J-447: Fixed bug for event available for patient and set error hand…

mhornfeldt:
…ler for xml validator to correct instance


mww-sopra-steria (approved):
Looks good! Nice to find the unnecessary instantiation on the error handler.

PR #337, opened 2024-11-04
K1J-481: When sending an answer to the recipient the answer should be set as 'HANDLED' instead of 'SENT'.

mww-sopra-steria:


ChrilleSigerhed (approved):
Nice, well done!

PR #338, opened 2024-11-04
K1J-483: Consider messages with answers or reminders when calculating unhandled message statistics

mww-sopra-steria:


ChrilleSigerhed (approved):
Very nice!

PR #339, opened 2024-11-04
K1J-467: Updated confirmation modal alert

mhornfeldt:


certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificatemodel/model/AlertType.java
```diff
@@ -0,0 +1,5 @@
+package se.inera.intyg.certificateservice.domain.certificatemodel.model;
+
+public enum AlertType {
+  INFO, OBSERVE, ERROR
```
mww-sopra-steria, line 4: There is something similar already in the model that is called ElementMessageLevel. Could we do something in order to use the same concept? At least within the model?

mww-sopra-steria (changes requested):
Looks good! Just one thought relating to AlertType to consider.

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/application/certificate/dto/AlertDTO.java
```diff
@@ -0,0 +1,22 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import lombok.Builder;
+import lombok.Value;
+import se.inera.intyg.certificateservice.application.certificate.dto.AlertDTO.AlertBuilder;
+import se.inera.intyg.certificateservice.domain.certificatemodel.model.MessageLevel;
+
+@JsonDeserialize(builder = AlertBuilder.class)
+@Value
+@Builder
+public class AlertDTO {
```
ChrilleSigerhed, line 13: I think we need to update the AlertBuilder to AlertDTOBuilder?

ChrilleSigerhed (approved):
Minor DTO update needed, but everything else looks good. Well done

PR #341, opened 2024-11-08
K1J-370: Added overflow pages for PDF

mhornfeldt:


mikaelaromild1 (approved):
Well done, approved! I will add comments/suggestions afterwards on what can be cleaned up and maybe improved at a later stage :)

PR #344, opened 2024-11-11
K1J-499: Midwife should be allowed to sign, revoke and handle protect…

mhornfeldt:
…ed person certificates for FK7472


mww-sopra-steria (approved):
Nice!

PR #346, opened 2024-11-11
K1J-495: Fixed incorrect publishing of messageId and updated logic to fetch correct message when building XML representation.

ChrilleSigerhed:


mww-sopra-steria (changes requested):
I found publishing the message id instead of answer id a bit unintuitive because the event is always "message-sent" no matter if it is an answer or question.

After reading the code in certificate-service and intygstjanst I'm suspicious that this is not the solution. I believe that the event should contain the answer id, but there is something wrong with the code that returns the xml-representation for the answer.

Best to investigate what the correct "amneskod" should be considering that we should send the answer to the recipient.

mww-sopra-steria (approved):
Looks great! 🙌

PR #347, opened 2024-11-12
K1J-499: Midwife should have signing rights for fk7472 and there for …

mhornfeldt:
…not forward cert or ready for sign


ChrilleSigerhed (approved):
Nice!

PR #355, opened 2024-12-04
K1J-560: Added TS8071

mhornfeldt:


certificate-service/app/src/main/java/se/inera/intyg/certificateservice/testability/certificate/service/fillservice/ts8071/TestabilityCertificateFillServiceTS8071.java
```diff
@@ -0,0 +1,269 @@
...
+import se.inera.intyg.certificateservice.domain.certificatemodel.model.ElementSpecification;
+import se.inera.intyg.certificateservice.domain.certificatemodel.model.FieldId;
+import se.inera.intyg.certificateservice.domain.common.model.Code;
+import se.inera.intyg.certificateservice.infrastructure.certificatemodel.common.codesystems.CodeSystemKnowledge;
+import se.inera.intyg.certificateservice.infrastructure.certificatemodel.common.codesystems.CodeSystemPositionHearingAid;
+import se.inera.intyg.certificateservice.testability.certificate.dto.TestabilityFillTypeDTO;
+import se.inera.intyg.certificateservice.testability.certificate.service.fillservice.TestabilityCertificateFillService;
+
+@Component
+public class TestabilityCertificateFillServiceTS8071 implements
```
mhornfeldt, line 104: If we like this format this can be made into a general solution if time in this sprint or in the future. I think it's easier to write it in this way rather than filling in each question with its own function

mww-sopra-steria (approved):
Great job! It is a huge certificate! :-)

PR #360, opened 2024-12-12
K1J-591: Implemented general print integration to certificate print service

mhornfeldt:


mhornfeldt:
Have not tested with CPS will add what is needed for serializing when able to test

certificate-service/integration-certificate-print-service/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+
+group = 'se.inera.intyg.certificateservice'
+version = '0.0.1-SNAPSHOT'
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    implementation 'org.apache.pdfbox:pdfbox:3.0.2'
```
mww-sopra-steria, line 13: Do we need to use pdf-box in this module? If we do then we shouldn't hardcode version.

certificate-service/integration-certificate-print-service/src/main/java/se/inera/intyg/certificateservice/certificate/integration/PrintCertificateFromCertificatePrintService.java
```diff
@@ -0,0 +1,49 @@
...
+    try {
+      return restClient
+          .post()
+          .uri(printCertificateServiceUrl + "/api/print/" + certificateId)
+          .header(LOG_TRACE_ID_HEADER, MDC.get(TRACE_ID_KEY))
+          .header(LOG_SESSION_ID_HEADER, MDC.get(SESSION_ID_KEY))
+          .body(request)
+          .retrieve()
+          .body(PrintCertificateResponseDTO.class);
+    } catch (HttpClientErrorException ex) {
```
mww-sopra-steria, line 42: Do we need to catch? Why return null if status is NOT_FOUND?

mww-sopra-steria (changes requested):
Looks really good! Nice work!

A few questions/comments to look at before merging.

certificate-service/integration-certificate-print-service/src/main/java/se/inera/intyg/certificateservice/certificate/integration/PrintCertificateFromCertificatePrintService.java
mhornfeldt: used same pattern as intygsadmin but can update

mww-sopra-steria: I don't think we should return null, suggest that we don't catch any exceptions.

mww-sopra-steria (changes requested):

mww-sopra-steria (approved):

PR #362, opened 2024-12-13
K1J-607: CertificateActionSend should handle multiple recipients.

ChrilleSigerhed:


certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/action/certificate/model/CertificateActionSend.java
```diff
@@ -42,17 +50,56 @@ public boolean evaluate(Optional<Certificate> certificate,
 
   @Override
   public String getName(Optional<Certificate> certificate) {
-    return NAME;
+    return certificate
+        .map(Certificate::certificateModel)
```
mhornfeldt, line 54: This is similar for all three could be general maybe with just specific text

mhornfeldt (approved):

PR #368, opened 2024-12-16
K1J-607: Refactor ElementValidation to enable validation depending on another field value.

ChrilleSigerhed:


certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/validation/model/ElementValidation.java
```diff
@@ -7,5 +7,6 @@
 
 public interface ElementValidation {
 
-  List<ValidationError> validate(ElementData data, Optional<ElementId> categoryId);
-}
+  List<ValidationError> validate(ElementData data, Optional<ElementId> categoryId,
+      List<ElementData> elementData);
```
mww-sopra-steria, line 11: Maybe change name of elementData to dataList to match the first argument that is data. Whatever name decided make sure that the implementations reflect the name.

mww-sopra-steria (approved):
Nice! Sorry to keep you waiting! 

Check comment before merge. I think it would be good to double check the naming of the new argument.

PR #373, opened 2024-12-17
K1J-625: Shall exclude categories if certificate is signed if the category does not contain any questions with element value.

ChrilleSigerhed:


mhornfeldt (approved):
nice!

PR #375, opened 2024-12-17
K1J-611: Implemented question correction of visual acuity with XML mapping.

ChrilleSigerhed:


certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/common/codesystems/CodeSystemKorringerAvSynskarpa.java
```diff
@@ -0,0 +1,31 @@
+package se.inera.intyg.certificateservice.infrastructure.certificatemodel.common.codesystems;
+
+import se.inera.intyg.certificateservice.domain.common.model.Code;
+
+public class CodeSystemKorringerAvSynskarpa {
```
mww-sopra-steria, line 5: Rename CodeSystemKorringerAvSynskarpa -> CodeSystemKorrigeringAvSynskarpa

mhornfeldt (approved):
Nice!

certificate-service/clinicalprocess-certificate-v4/src/main/java/se/inera/intyg/certificateservice/infrastructure/clinicalprocesscertificatev4/certificate/XmlGeneratorValue.java
```diff
@@ -77,7 +86,7 @@ private List<XmlMapping> mappings(CertificateModel certificateModel,
   }
 
   private static boolean noCustomMapping(XmlMapping mapping) {
-    return mapping.getMapping() == null;
+    return mapping.getMapping() == null || mapping.getMapping().elementValueMapper().isPresent();
```
mww-sopra-steria, line 89: I think it would be good if we changed the logic here a little bit. As it is written now we can't have custom mapping if we provide a elementValueMapper. But let's say we would have liked to have a custom mapper and a an elementValueMapperr?

Maybe use something like this instead: 
`return mapping.getMapping() == null || mapping.getMapping().elementId() == null;`

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificate/model/CodeListToBoolean.java
```diff
@@ -0,0 +1,14 @@
+package se.inera.intyg.certificateservice.domain.certificate.model;
+
+import lombok.Builder;
+import lombok.Value;
+
+@Value
+@Builder
+public class CodeListToBoolean implements ElementValue {
```
mww-sopra-steria, line 8: To handle the mapping as a "fictive" ElementValue is little of a "hack". But it would require some refactoring to get rid of it. I suggest we add a refactor-task that deals with this but leave it as implemented in this PR.

mww-sopra-steria (changes requested):
Nicely done! Good job!

A couple of minor adjustments + comment to add a refactoring task (reason is to be able to release this to STAGE and we can refactor later but during this sprint or the next..

mww-sopra-steria (approved):
Well done!

PR #409, opened 2025-02-26
K1J-161: Sent more data to WC to populate more events for admin info

mhornfeldt:


certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificate/model/CertificateMetaData.java
```diff
@@ -20,4 +20,5 @@ public class CertificateMetaData {
   CareUnit careUnit;
   CareProvider careProvider;
   ResponsibleIssuer responsibleIssuer;
+  Staff creator;
```
ChrilleSigerhed, line 23: Consider updating this to createdBy, so that it better matches revokedBy.

ChrilleSigerhed (approved):
Looks good!

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificate/model/CertificateMetaData.java
mhornfeldt: revokedBy isn't in this object. I was matching it against issuer. Should I leave it?

PR #437, opened 2025-04-15
K1J-970: Implemented component date range

mikaelaromild1:


certificate-service/clinicalprocess-certificate-v4/src/main/java/se/inera/intyg/certificateservice/infrastructure/clinicalprocesscertificatev4/certificate/XmlGeneratorDateRange.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    if (dateRangeValue.toDate() == null && dateRangeValue.fromDate() == null) {
+      return Collections.emptyList();
+    }
+
+    final var objectFactory = new ObjectFactory();
+
+    final var dateRangeAnswer = new Svar();
+    dateRangeAnswer.setId(
```
ChrilleSigerhed: We should always be able to use the id from data, so I think we can update this to

```java
  final var dateRangeAnswer = new Svar();
  dateRangeAnswer.setId(data.id().id());
```

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/certificatemodel/model/ElementType.java
```diff
@@ -3,5 +3,5 @@
 public enum ElementType {
   CATEGORY, DATE, ISSUING_UNIT, TEXT_AREA, CHECKBOX_DATE_RANGE_LIST, CHECKBOX_MULTIPLE_DATE,
   RADIO_MULTIPLE_CODE, RADIO_BOOLEAN, MESSAGE, DIAGNOSIS, TEXT_FIELD, MEDICAL_INVESTIGATION_LIST,
-  CHECKBOX_MULTIPLE_CODE, VISUAL_ACUITIES
-}
\ No newline at end of file
+  CHECKBOX_MULTIPLE_CODE, VISUAL_ACUITIES, PERIOD_DATE, DATE_RANGE
```
ChrilleSigerhed, line 6: PERIOD_DATE can be removed since its not used.

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/validation/model/ElementValidationDateRange.java
```diff
@@ -0,0 +1,140 @@
...
+  @Override
+  public List<ValidationError> validate(ElementData data,
+      Optional<ElementId> categoryId, List<ElementData> dataList) {
+    if (data == null) {
+      throw new IllegalArgumentException("Element data is null");
+    }
+
+    final var dateRange = getValue(data.value());
+
+    final var rangeValidationErrors = getRangeValidationErrors(data, categoryId, dateRange);
```
ChrilleSigerhed, line 35: Consider moving the ```rangeValidationErrors``` so that they are done after we made the mandatory check. I think that might help to simplify the logic added in ```getRangeValidationErrors``` and we can remove: 

```java
final var incompleteErrors = getIncompleteDateRangeErrors(data, categoryId, dateRange);
```

and instead just use

```java
private List<ValidationError> getToBeforeFromDateRangeErrors(ElementData data,
      Optional<ElementId> categoryId,
      ElementValueDateRange dateRange) {
    return isToBeforeFrom(dateRange) ?
        List.of(errorMessage(
                data, getFieldId(dateRange.id(), RANGE_SUFFIX), categoryId,
                ErrorMessageFactory.endDateAfterStartDate()
            )
        ) : Collections.emptyList();
  }
```

ChrilleSigerhed (changes requested):
Looks great, well done!

Left some comments to consider or maybe discuss before merging.

Also, should we update the schematron and add tests for it in this PR or do you think we should handle it in a separate PR?

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/validation/model/ElementValidationDateRange.java
mhornfeldt: We can't replace incompleteErrors with mandatory since the field ids are different.

ChrilleSigerhed: I see.

Would it be a good idea so separate the logic in ```getIncompleteDateRangeErrors``` and validate the ```getIncompleteDateRangeErrors``` in a separate method?

I just found ```getRangeValidationErrors``` alittle complex to grasp and I think that would help and make the code easier to follow and understand.

mhornfeldt: The logic is already separated and validated in separate methods

ChrilleSigerhed (approved):
Awsome!

PR #457, opened 2025-06-19
K1J-687: Added PDF support for FK3221.

ChrilleSigerhed:


certificate-service/app/src/test/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/fk3221/elements/QuestionOvrigtTest.java
```diff
@@ -64,4 +66,17 @@ void shallIncludeValidations() {
...
     assertEquals(expectedValidations, element.validations());
   }
-}
\ No newline at end of file
+
+  @Test
+  void shallIncludePdfConfiguration() {
+    final var expected = PdfConfigurationText.builder()
+        .pdfFieldId(new PdfFieldId("form1[0].#subform[3].flt_txtUpplysningar[0]"))
+        .maxLength(371)
```
mhornfeldt, line 74: Did you get the limits from FK? Otherwise for the past certificates we have used a fixed number for max char that fits on a row, place this in the specification and multiplied with the rows for each text area. If some elements don't have full rows (or longer rows than standard) it needs to be adjusted

mikeyOne (approved):

PR #475, opened 2025-07-02
K1J-1189: Validate sub answer id for questions when prefilling.

ChrilleSigerhed:


LucasLundstrom (changes requested):
We should not only verify that the subanswer with correct id exists but also filter that specific subanswer when getting the content.

LucasLundstrom (approved):
Nice work!

PR #479, opened 2025-07-08
K1J-1208: Fix code system mapping

LucasLundstrom:


ChrilleSigerhed (approved):
Nice! I assume we need to update the prefill aswell? But since they are disabled it does not matter for now. But might be worth updating in this PR?

PR #487, opened 2025-07-18
K1J-1222: enable integration test for prefill certificate

LucasLundstrom:


ChrilleSigerhed:
Added another solution that updates with gradle task [https://github.com/sklintyg/core/pull/490](url)

Maybe its to prefer since then we dont need to manually update dates to enable regex match.

mhornfeldt:
@ChrilleSigerhed should this be closed?

PR #493, opened 2025-08-05
K1J-1229: Add support for resource link FMB, SRS_FULL_VIEW & SRS_MINIMIZED_VIEW.

ChrilleSigerhed:


certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/user/model/User.java
```diff
@@ -18,15 +18,16 @@
...
+    Role role;
+    List<PaTitle> paTitles;
+    List<Speciality> specialities;
+    Blocked blocked;
+    Agreement agreement;
+    AllowCopy allowCopy;
+    AccessScope accessScope;
+    List<HealthCareProfessionalLicence> healthCareProfessionalLicence;
+    ResponsibleIssuer responsibleIssuer;
+    SrsActive srs;
```
mww-sopra-steria, line 32: Should we call the property srsActive to make it more clear? srs in itself doesn't say much what it is. Of course when you check the type SrsActive you understand... but it could make it more readable.

certificate-service/domain/src/main/java/se/inera/intyg/certificateservice/domain/user/model/SrsActive.java
```diff
@@ -0,0 +1,5 @@
+package se.inera.intyg.certificateservice.domain.user.model;
+
+public record SrsActive(boolean value) {
+
```
mww-sopra-steria, line 4: As we discussed, maybe we should set value to false if SrsActive is created with null. It could happen if a consumer for whatever reason don't provide a value for SrsActive when calling CS.

mww-sopra-steria (approved):
Looks good! Two minor comments to deal with before merge.

PR #496, opened 2025-08-06
K1J-1242: Implement FK7804 certificate send content provider to handle custom body.

ChrilleSigerhed:


mww-sopra-steria (approved):
Looks great!

PR #523, opened 2025-08-25
K1J-1244: Schematron fix

alryabov:
- Q27-Q29 are used multiple times in file. 
- Those are redundant, there are already assersions present, that are more complete and covers the same assertion.


ChrilleSigerhed (approved):
Nice, looks good!

Please update the schematron uploaded to confluence after merging.

PR #528, opened 2025-08-28
K1J-1305: Added prefill from xml to renew external certificate

mhornfeldt:


mww-sopra-steria (changes requested):
Looks really good! Nice to see that the functionality could be extended so well.

I have few minor suggestions that I believe will improve the readability:
1) Instead of overloading the prefill-method with a "boolean-version" take the opportunity to use the state of the MedicalCertificate domain object. It knows that it is and extension of another certificate (parent-relation of certain type) and based on that check if "includeWhenRenewing".
2) Remove the SubUnit argument from prefill-method and instead use the metadata inside the MedicalCertificate as it already knows which subunit it is related to (I know this was not added in this PR but a refactoring that in a way goes in line with bullet 1)
3) Remove the parent(...) method from Certificate interface and its implementation classes. (This wasn't added in this PR either but we should avoid exposing modifying the state of the object with setters)

PR #532, opened 2025-09-01
K1J-1315: Add placeholder certificate existence check and revocation endpoints.

ChrilleSigerhed:


mww-sopra-steria (commented):
I think everything looks good and I don't really have anything to comment on. 

But I wonder if we should simplify the implementation relating to placeholder-certificate. The implementation in this PR follows the pattern for MedicalCertificate, with the exception that the revoke is handled in an internal api. I would like to discuss if it is necessary or maybe even unwanted.

An alternative is to only have an internal api that revokes any existing placeholder. The API just handles the request and call the repository to revoke any existing placeholder. 

This will remove the need for exists calls, revoke implementation in domain object and save-logic. 

The placeholder logic is preferably kept as limited as possible. Could we discuss the option before deciding? But I'm really fine with both options.

PR #558, opened 2025-09-29
K1J-1416: Add unit of measurement to CertificateDataValue and update related tests. Updated configuration for QuestionKontakt.

ChrilleSigerhed:
Remove unitOfMeasurement from QuestionAntalManader, since it should not be displayed for this question.

Add unitOfMeasurement to value, since it should support it, even tho we are not using it atm (It's used in 1177)

Update config for QuestionKontakt so that correct values are displayed when the certificate is signed


mww-sopra-steria (approved):
Nice that you found the bug! Will save us in the future!

PR #568, opened 2025-10-03
K1J-1432: Handle messages with relation and recipient

mww-sopra-steria:
Added implementation to handle additional properties in the analytics message:
1. recipient - used on events that send the certificate
2. relation - used on certificates that has a parent relationship to another certificate

Persists the recipient in a new dimension table.

Persists the relation in new dimension tables connected to certificate.

Integration tests to verify that the additions can be handled correctly from message to persist.


ChrilleSigerhed (approved):
Looks good, well done!

PR #569, opened 2025-10-03
K1J-1433: Added administrative question fields to event messages

mhornfeldt:


certificate-analytics-service/app/src/main/java/se/inera/intyg/certificateanalyticsservice/application/messages/model/v1/CertificateAnalyticsEventMessageV1.java
```diff
@@ -19,6 +19,7 @@ public class CertificateAnalyticsEventMessageV1 implements CertificateAnalyticsM
   CertificateAnalyticsEventCertificateV1 certificate;
   CertificateAnalyticsEventV1 event;
   CertificateAnalyticsEventRecipientV1 recipient;
+  CertificateAnalyticsAdministrativeMessageV1 administrativeMessage;
```
mhornfeldt, line 22: Not sure about the name but if we only name it message it gets confusing to message for the event

certificate-analytics-service/app/src/main/java/se/inera/intyg/certificateanalyticsservice/infrastructure/persistance/entity/AdministrativeMessageEntity.java
```diff
@@ -0,0 +1,70 @@
...
+  @JoinColumn(name = "administrative_message_type_key", referencedColumnName = "key", nullable = false)
+  private AdministrativeMessageTypeEntity messageType;
+
+  @Column(name = "sent")
+  private LocalDateTime sent;
+
+  @Column(name = "last_date_to_answer")
+  private LocalDate lastDateToAnswer;
+
+  @ElementCollection(fetch = FetchType.EAGER)
```
mhornfeldt, line 55: Suggested by Copilot for efficient storing lists. We could have many to many otherwise I guess

mhornfeldt:
Asked question about if we want administrative message entity to update when new events come (reminder added, answer added etc.) need to update with this once verified

certificate-analytics-service/app/src/main/java/se/inera/intyg/certificateanalyticsservice/application/messages/model/v1/CertificateAnalyticsEventMessageV1.java
mww-sopra-steria: I agree it gets confusing. I do think it is better to make some adjustments on naming of the other classes so we can call the attribute message instead. The message parts can be both administrative and complement message even though both could be considered "administrative".

mww-sopra-steria:
I continued working on adding message and because it changed to a large extent I decided to close this draft pr and instead create a new.

certificate-analytics-service/app/src/main/java/se/inera/intyg/certificateanalyticsservice/infrastructure/persistance/entity/AdministrativeMessageEntity.java
mhornfeldt: This was a draft PR. I wrote the comment because I wanted to remind myself I needed to verify it. We have agreed before that the usage of AI tools should be transparant within the team. Highlighting it was a solution I wasn't sure about when adding it to the draft and then looking into it follows this pattern. If I would've added an AI-label and suggesting it to be in the final code, it would not have followed the workflow.

PR #574, opened 2025-10-07
K1J-1203: Added model for AG1-14

mhornfeldt:


mhornfeldt:
Includes adding certificate AG1-14 2.0. No new functionality introduced. Print function is blocked by K1J-1202.

certificate-service/app/src/main/java/se/inera/intyg/certificateservice/infrastructure/certificatemodel/ag114/elements/QuestionFormedlaDiagnos.java
```diff
@@ -0,0 +1,51 @@
...
+                    "Önskar patienten förmedla information om diagnos/diagnoser till sin arbetsgivare?")
+                .description(
+                    "Information om diagnos kan vara viktig för patientens arbetsgivare. Det kan underlätta anpassning av patientens arbetssituation. Det kan också göra att patienten snabbare kommer tillbaka till arbetet.")
+                .selectedText("Ja")
+                .unselectedText("Nej")
+                .build()
+        )
+        .rules(
+            List.of(
+                CertificateElementRuleFactory.mandatory(
```
mww-sopra-steria, line 36: This rule should be modified to mandatoryOrExist(...). Now it looks like this:

Image

mww-sopra-steria: I noticed we have the same issue in more places so it would be good to double check the mandatory rules.

